PR #23, opened 2020-02-20
Stacksafe version of lazyAp

### LINKED ISSUE #7: ["Bug"] Traverse on infinite sequences can stackoverflow
```kotlin
generateSequence { 0 }.k().parTraverse(SequenceK.traverse()) {
        IO { println(it) }
    }.unsafeRunSync()
// => stackoverflow
```
While this code is nonsense anyway because it will infinite loop without ever printing values, it should not stackoverflow and it shows that `traverse` is currently not stacksafe!

This is caused by poor use of `Eval`:
https://github.com/arrow-kt/arrow/blob/712d7cf917cbf60958d1918c113bbc24b22c70f8/modules/core/arrow-core-data/src/main/kotlin/arrow/core/SequenceK.kt#L27
And similar: 
https://github.com/arrow-kt/arrow/blob/712d7cf917cbf60958d1918c113bbc24b22c70f8/modules/core/arrow-core-data/src/main/kotlin/arrow/core/ListK.kt#L160

I honestly don't know how to fix this. This will only become a problem if used with a stackunsafe applicative (Either, Option e.g.) When used on something that is stacksafe and implements lazyAp properly like the normal IO applicative everything will work fine.

1Jajen1:
Uses `Eval` to defer computation instead of a simple function. Fixes #7 and later on when this is added to fx https://github.com/arrow-kt/arrow-fx/issues/7


nomisRev:
CI is 💚

nomisRev:
And it was a false positive due to the re-org 😭

1Jajen1:
Those are `AbstractMethodError`'s :scream:

1Jajen1:
Wait, can this be caused by the test using `Id.applicative` from the current snapshot version and thus not having the `lazyAp` changes from this branch?! @rachelcarmena ideas?

1Jajen1:
> Wait, can this be caused by the test using `Id.applicative` from the current snapshot version and thus not having the `lazyAp` changes from this branch?! @rachelcarmena ideas?

Given that only the tests that use `lazyAp` through an applicative instance not provided from `arrow-core` directly (`TraverseLaws` uses `Id.applicative` which comes from `arrow-core` but likely the compiled snapshot `arrow-test` has and not this branch) I think that this may very well be the problem.

1Jajen1:
A solution may be what I talked about in https://github.com/arrow-kt/arrow-test/pull/7#issuecomment-589599272. Splitting `arrow-test` and have one for each arrow-repo. This would prevent most of these issues...

arrow-core-data/src/main/kotlin/arrow/typeclasses/Apply.kt
```diff
@@ -38,11 +38,10 @@ interface Apply<F> : Functor<F> {
    */
   fun <A, B> Kind<F, A>.ap(ff: Kind<F, (A) -> B>): Kind<F, B>
 
-  /**
-   * Lazy version of ap, useful for datatypes which can short circuit.
-   * This will at some point be removed for a compiler plugin
-   */
-  fun <A, B> Kind<F, A>.lazyAp(ff: () -> Kind<F, (A) -> B>): Kind<F, B> = ap(ff())
+  fun <A, B> Kind<F, A>.lazyAp(ff: Eval<Kind<F, (A) -> B>>): Eval<Kind<F, B>> = ff.map { this.ap(it) }
```
nomisRev: Like discussed on Slack.

Since this function is now not _lazy_ anymore but instead uses `Eval` we should rename to `apEval` which is fine since `apLazy` has not been released yet. Amazing work @1Jajen1 👏 👏

arrow-core-test/src/main/kotlin/arrow/test/laws/ComonadLaws.kt
```diff
@@ -22,9 +21,7 @@ object ComonadLaws {
         Law("Comonad Laws: duplicate then map into extract is identity") { CM.duplicateThenMapExtractIsId(GEN, EQ) },
         Law("Comonad Laws: map and coflatMap are coherent") { CM.mapAndCoflatmapCoherence(GEN, EQ) },
         Law("Comonad Laws: left identity") { CM.comonadLeftIdentity(GEN, EQ) },
-        Law("Comonad Laws: right identity") { CM.comonadRightIdentity(GEN, EQ) },
-        Law("Comonad Laws: cokleisli left identity") { CM.cokleisliLeftIdentity(GEN, EQ) },
-        Law("Comonad Laws: cokleisli right identity") { CM.cokleisliRightIdentity(GEN, EQ) }
```
1Jajen1: These are alternative definitions only added for parity with monad laws. They follow directly from the other laws and are thus not needed

arrow-core-test/src/main/kotlin/arrow/test/laws/MonadLaws.kt
```diff
@@ -33,8 +32,6 @@ object MonadLaws {
       listOf(
         Law("Monad Laws: left identity") { M.leftIdentity(G, EQ) },
         Law("Monad Laws: right identity") { M.rightIdentity(G, EQ) },
-        Law("Monad Laws: kleisli left identity") { M.kleisliLeftIdentity(G, EQ) },
-        Law("Monad Laws: kleisli right identity") { M.kleisliRightIdentity(G, EQ) },
```
1Jajen1: These laws are alternatives to the left and right identity laws. They exist because they show the similarity between monoid laws but are itself not useful when the other two identity laws are in use

arrow-core-data/src/test/kotlin/arrow/typeclasses/TraverseTest.kt
```diff
@@ -0,0 +1,17 @@
...
+import arrow.core.Some
+import arrow.core.extensions.option.applicative.applicative
+import arrow.core.extensions.sequence.traverse.sequence
+import io.kotlintest.shouldBe
+import io.kotlintest.specs.StringSpec
+
+class TraverseTest : StringSpec({
+  "traverse is stacksafe over very long collections and short circuits properly" {
+    // This has to traverse 50k elements till it reaches None and terminates
+    generateSequence(0) { it + 1 }.map { if (it < 50_000) Some(it) else None }
```
nomisRev: Should this be a toggle-able `TraverseLaw` like we do for `ConcurrentLaws` and `MonadDeferLaws` with `testStacksafety` ?

1Jajen1: Don't think that is wise because this tests the applicative instance more than the traverse instance. This could be a toggleable law in apply which makes sure that `apEval` is stacksafe.

1Jajen1: But a toggle-able law in apply means a lot of laws need to add a default parameter of sorts to their constructors and apply has a lot of laws above it

1Jajen1: Doesn't matter too much too me anyway, but we could also make it non-toggleable because it uses `Eval.map` it should be stacksafe anyway, without a manual implementation it just won't short-circuit

1Jajen1:
Well the documentation build failure is for the exact same reason the tests failed before. Don't think I can do anything there. The build itself is now successful because I removed the last remaining references to fx and mtl.

rachelcarmena:
> Well the documentation build failure is for the exact same reason the tests failed before. Don't think I can do anything there. The build itself is now successful because I removed the last remaining references to fx and mtl.

I'm "cleaning" that part now :wink:

rachelcarmena:
> > Well the documentation build failure is for the exact same reason the tests failed before. Don't think I can do anything there. The build itself is now successful because I removed the last remaining references to fx and mtl.
> 
> I'm "cleaning" that part now wink

Done: #42

1Jajen1:
Since this should be a green build now and I renamed `lazyAp` -> `apEval` I think this is good to go.

rachelcarmena:
> Since this should be a green build now and I renamed `lazyAp` -> `apEval` I think this is good to go.

@nomisRev is doing a last review right now :wink: Thanks Simon!!

arrow-core-data/src/test/kotlin/arrow/typeclasses/TraverseTest.kt
nomisRev: Alright, makes sense!

1Jajen1:
> Great work!!! Awesome to have these changes finally in

Woo, time to merge then and do the follow up's for all the other repos :)

PR #24, opened 2020-02-20
Flipped argument version of ap

1Jajen1:
See https://github.com/arrow-kt/arrow/issues/1000 for more details. This does not make `ap` infix. Not sure if that'd be necessary, but that'd be an easy thing to add.

I'll prepare another branch adding a deprecation note to `ap` in a day or two and then mark this as ready.

Btw is `release/0.11.0` still the right target?


nomisRev:
> Btw is release/0.11.0 still the right target?

@rachelcarmena 
We can keep using the `release/0.11.0` in all repos, correct?

rachelcarmena:
> > Btw is release/0.11.0 still the right target?
> 
> @rachelcarmena
> We can keep using the `release/0.11.0` in all repos, correct?

Right, in all repos :+1: Branch protection rule and sync bot for all the repos.

1Jajen1:
Given that it has been a while and there have been many changes I'll reopen this with a new pr. I tried merging this but at some point it felt like more work than redoing it. (Which is fine cause this is just reversing arguments anyway...)

Issue #5, opened 2020-01-30
["Request"] Disambiguate `unit` function in Applicative/Functor to `void`


aballano:
We have a duplicated function name that performs 2 completely different things

unit for Applicative
`fun unit(): Kind<F, Unit> = just(Unit)`

and unit for Functor
`fun <A> Kind<F, A>.unit(): Kind<F, Unit> = map { Unit }`

The problem with this can be easily spotted when having an extension function over an interface extending, for example, MonadError, where you wanna start calling unit(), as both functions can be resolved, there's no guarantee of which one will be called, causing possible bugs. Here unit could either mean `just(Unit)` or `this.map { Unit }`:

```
fun <A> Kind<F, A>.something(...): Kind<F, A> =
    unit().doSomethingElse()
```

I'd suggest removing/renaming the unit() from Functor as the name is not indicating a map transformation whatsoever. Maybe simply `mapUnit`? cc @arrow-kt/maintainers 

EDIT: `Functor#void` should be renamed to `void` which is similar to Haskell and implies that the output is mapped to `Unit` or void'ed rather than returning a lifted `Unit` value from `Applicative`.

Since this is a breaking change this should be deprecated in 0.10.x and fully replaced in 0.11.0.

PR #30, opened 2020-02-21
Fix Eval's Fx not being lazy + deprecate Functor's unit

aballano:
From arrow-kt/arrow#2089

Added test to verify that the fx block runs lazily + fix for it
Nit: Adjust code style of Eval, Alternative, Functor and others


arrow-core-data/src/main/kotlin/arrow/typeclasses/Alternative.kt
```diff
@@ -80,11 +87,17 @@ fun <T, F, A> Kind<T, Kind<F, A>>.altSum(AF: Alternative<F>, FT: Foldable<T>): K
...
-fun <T, F, A> Kind<T, A>.altFold(AF: Alternative<F>, FT: Foldable<T>): Kind<F, A> = FT.run { toList().altFromList(AF) }
+fun <T, F, A> Kind<T, A>.altFold(AF: Alternative<F>, FT: Foldable<T>): Kind<F, A> =
+  FT.run { toList().altFromList(AF) }
 
-fun <F, A> List<A>.altFromList(AF: Alternative<F>): Kind<F, A> = map { AF.just(it) }.k().altSum(AF, object : Foldable<ForListK> {
-  override fun <A, B> Kind<ForListK, A>.foldLeft(b: B, f: (B, A) -> B): B = fix().foldLeft(b, f)
-  override fun <A, B> Kind<ForListK, A>.foldRight(lb: Eval<B>, f: (A, Eval<B>) -> Eval<B>): Eval<B> = fix().foldRight(lb, f)
-})
+fun <F, A> List<A>.altFromList(AF: Alternative<F>): Kind<F, A> =
+  map { AF.just(it) }.k().altSum(AF, object : Foldable<ForListK> {
```
nomisRev: Can we not use `ListK.foldable()` here directly? Instead of re-creating the `Foldable<ForListK>` object every time we call this function?

arrow-core-data/src/main/kotlin/arrow/typeclasses/Functor.kt
```diff
@@ -130,7 +132,8 @@ interface Functor<F> : Invariant<F> {
    * }
    * ```
    */
-  fun <A> Kind<F, A>.unit(): Kind<F, Unit> = map { Unit }
+  fun <A> Kind<F, A>.unit(): Kind<F, Unit> =
```
pakoito: Should be changed in favor of `void`, now that we're here :D

aballano: Classic "poyaque" 😏

arrow-core-data/src/main/kotlin/arrow/typeclasses/Alternative.kt
1Jajen1: Not from `arrow-core-data` because `ListK.foldable` is in `arrow-core`. I wrote this and was very annoyed at that as well :/

arrow-core-data/src/main/kotlin/arrow/typeclasses/Functor.kt
1Jajen1: Btw should we deprecate this first and add a new one `void` or just go with a rename?

arrow-core-data/src/main/kotlin/arrow/typeclasses/Alternative.kt
nomisRev: Oh right, yes I've been there before :D Can we just keep a locally cached value then instead?

arrow-core-data/src/main/kotlin/arrow/typeclasses/Functor.kt
aballano: I was about to ask the same :D cc @pakoito @nomisRev

nomisRev: We should probably deprecate and add `void`. We'll go through a short deprecation cycle since we're trying to release `0.11.0` asap anyway.

aballano:
Also fixes https://github.com/arrow-kt/arrow-core/issues/5

PR #32, opened 2020-02-21
Change firstOption implementation

danimontoya:
This PR approaches the ticket [#22](https://github.com/arrow-kt/arrow-core/issues/22)

Changed firstOption implementation to be more like stdlib `firstOrNone`.

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/typeclasses/Foldable.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/typeclasses/Foldable.kt
@@ -205,18 +205,14 @@ interface Foldable<F> {
   /**
    * Get the first element of the foldable or none
    */
-  fun <A> Kind<F, A>.firstOption(): Option<A> = get(0)
+  fun <A> Kind<F, A>.firstOption(): Option<A> =
+    find { true }
 
   /**
    * Get the first element of the foldable or none if empty or the predicate does not match
    */
   fun <A> Kind<F, A>.firstOption(predicate: (A) -> Boolean): Option<A> =
-    foldLeft<A, Either<A, Long>>(0L.right()) { i, a ->
-      i.flatMap {
-        if (predicate(a)) Left(a)
-        else Right(it + 1L)
-      }
-    }.swap().toOption()
+    find { predicate(it) }
 
   fun <A> Kind<F, A>.toList(): List<A> = foldRight(Eval.now(emptyList<A>())) { v, acc -> acc.map { listOf(v) + it } }.value()
 
--- a/arrow-core-data/src/main/kotlin/arrow/typeclasses/Foldable.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/typeclasses/Foldable.kt
@@ -211,16 +211,21 @@ interface Foldable<F> {
    * Get the first element of the foldable or none if empty or the predicate does not match
    */
   fun <A> Kind<F, A>.firstOption(predicate: (A) -> Boolean): Option<A> =
-    get(0).filter(predicate)
+    foldLeft<A, Either<A, Long>>(0L.right()) { i, a ->
+      i.flatMap {
+        if (predicate(a)) Left(a)
+        else Right(it + 1L)
+      }
+    }.swap().toOption()
 
   fun <A> Kind<F, A>.toList(): List<A> = foldRight(Eval.now(emptyList<A>())) { v, acc -> acc.map { listOf(v) + it } }.value()
 
   companion object {
     @Deprecated("This function will be removed soon. Use Iterator.iterateRight from Eval.kt instead")
     fun <A, B> iterateRight(it: Iterator<A>, lb: Eval<B>): (f: (A, Eval<B>) -> Eval<B>) -> Eval<B> = { f: (A, Eval<B>) -> Eval<B> ->
-        fun loop(): Eval<B> =
-          Eval.defer { if (it.hasNext()) f(it.next(), loop()) else lb }
-        loop()
-      }
+      fun loop(): Eval<B> =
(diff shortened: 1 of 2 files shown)


arrow-core-data/src/main/kotlin/arrow/typeclasses/Foldable.kt
```diff
@@ -211,16 +211,21 @@ interface Foldable<F> {
    * Get the first element of the foldable or none if empty or the predicate does not match
    */
   fun <A> Kind<F, A>.firstOption(predicate: (A) -> Boolean): Option<A> =
-    get(0).filter(predicate)
+    foldLeft<A, Either<A, Long>>(0L.right()) { i, a ->
+      i.flatMap {
+        if (predicate(a)) Left(a)
+        else Right(it + 1L)
+      }
+    }.swap().toOption()
```
1Jajen1: While this works it's not great for very large structures, especially not lazy ones like sequence because `foldLeft` won't stop after it found something.

Can you redo this with `foldRight`? It should be possible to just write something like:
```kotlin
foldRight(Eval.now(None)) { x, evalAcc ->
  if (p(x)) Eval.now(x.some())
  else evalAcc
}
```

danimontoya: I see! I'll change it!

danimontoya: Although isn't that the exact same impl of `find` ?

1Jajen1: > Although isn't that the exact same impl of `find` ?

True, did not know that existed. It'll still be good to match the stdlib with `firstOption` I think. So defining it as an alias might not bee too bad. The stdlib has `find` as well which by the looks of it does the same thing as `firstOrNull`. 

I think doing `firstOption(p) == find(p)` will work fine. Oh and please change `firstOption() == find { true }`. The less we use `get` the better. Then we can get rid of the explicit overrides for `firstOption` here: https://github.com/arrow-kt/arrow-core/blob/009099e6c07312b60188dadf7b9ea84d831eea32/arrow-core/src/main/kotlin/arrow/core/extensions/sequenceK.kt#L166-L180

danimontoya: @1Jajen1 thanks for the explanation!! :D

1Jajen1: Thanks for changing this^^ This has been annoying me for a long time :)

arrow-core-data/src/main/kotlin/arrow/typeclasses/Foldable.kt
```diff
@@ -205,22 +205,23 @@ interface Foldable<F> {
   /**
    * Get the first element of the foldable or none
    */
-  fun <A> Kind<F, A>.firstOption(): Option<A> = get(0)
+  fun <A> Kind<F, A>.firstOption(): Option<A> =
+    find { true }
 
   /**
    * Get the first element of the foldable or none if empty or the predicate does not match
    */
```
1Jajen1: We should probably change this text

danimontoya: Something like..
_Get the first element matching the predicate or none_

1Jajen1: Sounds good

arrow-core-data/src/main/kotlin/arrow/typeclasses/Foldable.kt
```diff
@@ -205,22 +205,23 @@ interface Foldable<F> {
...
-  fun <A> Kind<F, A>.firstOption(): Option<A> = get(0)
+  fun <A> Kind<F, A>.firstOption(): Option<A> =
+    find { true }
 
   /**
-   * Get the first element of the foldable or none if empty or the predicate does not match
+   * Get the first element matching the predicate or none
    */
   fun <A> Kind<F, A>.firstOption(predicate: (A) -> Boolean): Option<A> =
```
aballano: btw, don't we have a test to verify this? IMO either the tests should either have failed or we should change them to verify this

1Jajen1: `Foldable` does have two tests but they only concern the two cases `F` is emtpy and the predicate always returns false. This does not cover the multiple elements match the predicate case.

danimontoya: Should we merge this change then and wait for the new split of `arrow-test` to extend the laws?
I've created a [new issue for it](https://github.com/arrow-kt/arrow-core/issues/34)

1Jajen1: I'm fine with merging this as is. Is `find` covered by any tests? If so, the missing test case won't matter until we think about covering derived functionality with tests anyway.

danimontoya: Sort of.. there is one law to check wether `exists` is consistent with `find`.. so definitely `FoldableLaws` test suite needs to be extended

1Jajen1: Hmm ok... I'm fine with merging anyway, as this is quite a bit more useful than before^^

`FoldableLaws` definitely needs work, quite a bit actually because with `Foldable` local overrides in specific instances (think specific search implementations for trees etc) make a ton of sense.
I am starting to dislike the term `Foldable*Laws*` though, because these are more tests than laws, but thats nitpicking and doesn't matter really :sweat_smile:

PR #36, opened 2020-02-24
Unify `Foldable` to be more aligned with Kotlin stdlib

### LINKED ISSUE #4: ["Request"] Unify `Foldable` API with Kotlin's language
In the `0.10.x` deprecation cycle we can take the opportunity to make the `Foldable` API more Kotlin idiomatic. We differ in quite some function names from the typical Kotlin naming.

 - [ ] `nonEmpty` vs `isNotEmpty`
 - [ ] `forAll` vs `all`
 - [ ] `find` vs `first`

danimontoya:
Fixes #4 .
All suggestions are welcome!
Deprecation of:
- forAll -> all
- nonEmpty -> isNotEmpty
- firstOption -> first
- firstOption(predicate) -> firstOrNone(predicate)

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/typeclasses/Foldable.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/typeclasses/Foldable.kt
@@ -147,7 +147,16 @@ interface Foldable<F> {
    *
    * If there are no elements, the result is true.
    */
+  @Deprecated("In favor of having a more Kotlin idiomatic API", ReplaceWith("all(p)"))
   fun <A> Kind<F, A>.forAll(p: (A) -> Boolean): Boolean =
+    all(p)
+
+  /**
+   * Check whether all elements satisfy the predicate.
+   *
+   * If there are no elements, the result is true.
+   */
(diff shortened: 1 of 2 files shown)


arrow-core-data/src/main/kotlin/arrow/typeclasses/Foldable.kt
```diff
@@ -205,13 +218,27 @@ interface Foldable<F> {
...
    * Get the first element matching the predicate or none
    */
+  @Deprecated("In favor of having a more Kotlin idiomatic API", ReplaceWith("firstOrNone(predicate)"))
   fun <A> Kind<F, A>.firstOption(predicate: (A) -> Boolean): Option<A> =
+    firstOrNone { predicate(it) }
+
+  /**
+   * Get the first element of the foldable or none
+   */
+  fun <A> Kind<F, A>.first(): Option<A> =
```
1Jajen1: imo this should be `firstOrNone` and not `first`

danimontoya: Changed it! thanks! 👍

PR #40, opened 2020-02-25
Add deprecation notice to ap

### LINKED ISSUE #6: Deprecate ap argument order
In arrow-kt/arrow#1898 ap parameter order was changed. This should be deprecated in 0.10.x

1Jajen1:
Fixes #6 

This is currently way too long imo. But I don't really know how to shorten that properly.

The replaceWith actually works :tada: so thats at least something. I'll keep this as a draft as long as #24 is not accepted and ready to merge.


1Jajen1:
Eww, I just noticed there is another annoying problem: `apEval`...
It will also change its signature with `ap`. This means either of two things: Flip it in `0.10.5` or add the same note to it and flip it in `0.11`. I'd probably go with the note because having two methods called `ap` with so different signatures seems weird.

nomisRev:
>  I'd probably go with the note because having two methods called ap with so different signatures seems weird.

+ 1

1Jajen1:
> I'll keep this as a draft as long as #24 is not accepted and ready to merge.

Actually forget this. If this blocks a 0.10.5 release just merge it at will. The flipped argument order pr is going to be another multi repo pr and thus will take a while...

Issue #34, opened 2020-02-24
["Request"] Extend FoldableLaws


danimontoya:
Foldable does have two tests covering `firstOption` but they only concern the two cases `F` is empty and the predicate always returns false.

This does not cover the multiple elements match the predicate case so we would have to extend the laws to cover this case.

Blocker: Split of `arrow-test` into the correspondent repos (so in this case, FoldableLaws should be part of `arrow-core`)
Needed to complete #32

PR #41, opened 2020-02-25
Extend test suite for FoldableLaws

danimontoya:
This PR extends the unit tests for Foldable to approach #34 .
Add unit tests for:
- [x] `foldRight` lazy
- [x] `find`
- [x] `firstOrNone`
- [x] `toList`
- [x] `fold`
- [x] `reduce`-*
- [x] `combineAll`
- [x] `orEmpty`
- [x] `traverse_` and `sequence_`
- [x] `isEmpty`
- [x] `isNotEmpty`
- [x] `foldMapM`
- [x] `get`


arrow-core-data/src/test/kotlin/arrow/core/ListKTest.kt
```diff
@@ -73,6 +74,7 @@ class ListKTest : UnitSpec() {
         this::bijection),
       MonoidKLaws.laws(ListK.monoidK(), ListK.genK(), ListK.eqK()),
       TraverseLaws.laws(ListK.traverse(), ListK.genK(), ListK.eqK()),
+      FoldableLaws.laws(ListK.foldable(), ListK.genK()),
```
1Jajen1: Isn't this already run through `TraverseLaws`?

danimontoya:
Am I missing more test cases?

arrow-core-test/src/main/kotlin/arrow/test/laws/FoldableLaws.kt
```diff
@@ -114,4 +130,31 @@ object FoldableLaws {
...
+  fun <F> Foldable<F>.`firstOrNone is consistent with find predicate`(G: Gen<Kind<F, Int>>, EQ: Eq<Option<Int>>) =
+    forAll(Gen.intPredicate(), G) { f: (Int) -> Boolean, fa: Kind<F, Int> ->
+      fa.firstOrNone(f).equalUnderTheLaw(fa.find { f(it) }, EQ)
+    }
+
+  fun <F> Foldable<F>.`toList turn items into a list`(G: Gen<Kind<F, Int>>, EQ: Eq<Int>) =
+    forAll(G) { fa: Kind<F, Int> ->
+      val expected = fa.foldRight(Eval.now(emptyList<Int>())) { v, acc -> acc.map { listOf(v) + it } }.value()
+      fa.toList().size.equalUnderTheLaw(expected.size, EQ)
+    }
```
1Jajen1: Just check both lists if they are equal. Only checking the size might not uncover ordering problem for example.

1Jajen1:
> Am I missing more test cases?

Likely, `Foldable` has quite the api. I'll check again

arrow-core-data/src/test/kotlin/arrow/core/ListKTest.kt
danimontoya: Yes, my fault!

1Jajen1:
Methods not covered yet:
- `fold`
- `reduce`-*
- `combineAll`
- `orEmpty` (although this method confuses me)
- `traverse_` and `sequence_`
- `isEmpty`
- `isNotEmpty`
- `foldMapM`
- `get`

Don't think this all needs to be done in one pr, but it's important to test^^

arrow-core-test/src/main/kotlin/arrow/test/laws/FoldableLaws.kt
danimontoya: I was struggling to get `Eq<List<Int>>`

1Jajen1: Two choices:
- Use `ListK.eq` and lift using `k`
- Use `Eq.any` both are trivial types hence `==` works just fine

arrow-core-test/src/main/kotlin/arrow/test/generators/Generators.kt
```diff
@@ -193,3 +198,8 @@ fun <A, B> Gen.Companion.genConst(gen: Gen<A>): Gen<Const<A, B>> =
   gen.map {
     Const<A, B>(it)
   }
+
+fun <A> Gen.Companion.genEval(gen: Gen<A>): Gen<Eval<A>> =
+  gen.map {
+    Eval.just(it)
+  }
```
1Jajen1: Might be nicer with `Gen<A>` as a receiver instead of a param.

arrow-core-test/src/main/kotlin/arrow/test/laws/FoldableLaws.kt
```diff
@@ -43,7 +55,26 @@ object FoldableLaws {
...
+  fun <F> laws(FF: Foldable<F>, GA: Applicative<F>, GENK: GenK<F>, EQK: EqK<F>): List<Law> {
+    val GEN: Gen<Kind<F, Int>> = GENK.genK(Gen.intSmall())
+
+    val EQKInt: Eq<Kind<F, Int>> = EQK.liftEq(Int.eq())
+    val EQKUnit = EQK.liftEq(Eq.any())
+
+    return listOf(
+      Law("Foldable Laws: orEmpty consistent with just empty") { FF.`orEmpty consistent with just empty`(GA, EQKInt) },
+      Law("Foldable Laws: traverse_ consistent with foldRight") { FF.`traverse_ consistent with foldRight`(GA, GENK, GEN, EQKUnit) },
+      Law("Foldable Laws: sequence_ consistent with traverse_") { FF.`sequence_ consistent with traverse_`(GA, GENK, EQKUnit) }
```
1Jajen1: I am against adding the applicative to the params. `traverse_` doesn't care about what instance is used, so just choose one for the tests. `Id` `Option` etc all are decent choices.

arrow-core-test/src/main/kotlin/arrow/test/laws/FoldableLaws.kt
```diff
@@ -154,7 +185,86 @@ object FoldableLaws {
...
-      fa.toList().size.equalUnderTheLaw(expected.size, EQ)
+      result.eqK(expected, EQ)
+    }
+
+  fun <F> Foldable<F>.`fold should combine all items`(G: Gen<Kind<F, Int>>, EQ: Eq<Int>) =
+    forAll(G) { fa: Kind<F, Int> ->
+      with(Int.monoid()) {
+        fa.fold(this).equalUnderTheLaw(fa.foldLeft(empty()) { acc, a -> acc.combine(a) }, EQ)
+      }
+    }
```
1Jajen1: Please use a non-commutative monoid here. That'd verify the correct order as well. Otherwise one could implement `fold` reversing the order which would never show with this test.

arrow-core-test/src/main/kotlin/arrow/test/laws/FoldableLaws.kt
```diff
@@ -154,7 +185,86 @@ object FoldableLaws {
...
+
+  fun <F> Foldable<F>.`combineAll consistent with fold (alias)`(G: Gen<Kind<F, Int>>, EQ: Eq<Int>) =
+    forAll(G) { fa: Kind<F, Int> ->
+      with(Int.monoid()) {
```
1Jajen1: Same as above

arrow-core-test/src/main/kotlin/arrow/test/laws/FoldableLaws.kt
danimontoya: I've pushed it to ask because I didn't know how to proceed there!! thanks for your fast response!! :D

arrow-core-test/src/main/kotlin/arrow/test/laws/FoldableLaws.kt
```diff
@@ -35,44 +44,47 @@ object FoldableLaws {
...
-      Law("Foldable Laws: fold returns combination of all items") { FF.`fold should combine all items`(GEN, EQ) },
-      Law("Foldable Laws: combineAll is an alias for fold") { FF.`combineAll consistent with fold (alias)`(GEN, EQ) },
-      Law("Foldable Laws: reduceLeftToOption combines all items into an optional value") { FF.`reduceLeftToOption returns Option value`(GENB, EQOptionInt) },
-      Law("Foldable Laws: reduceRightToOption combines all items into an optional value") { FF.`reduceRightToOption returns Option value`(GENB, EQOptionInt) },
-      Law("Foldable Laws: reduceLeftOption consistent with reduceLeftToOption") { FF.`reduceLeftOption returns Option value`(GEN, EQOptionInt) },
-      Law("Foldable Laws: reduceRightOption consistent with reduceRightToOption") { FF.`reduceRightOption returns Option value`(GEN, EQOptionInt) }
-    )
+    val EQListKInt = ListK.eq(Int.eq())
+
+    return laws(Option.traverseFilter(), Option.applicative(), Option.genK(), Option.eqK()) +
```
1Jajen1: hmm. Actually I was wrong. We need the overload that you had in the previous commit. But only for the `orEmpty` test. Do not use this to test anything that works over arbitrary `Applicative` instances (like `traverse_` and `sequence_`). This way we have `orEmpty` tested whenever our type is both `Foldable` and `Applicative` and we test everything else in the other cases. Oversight on my part sorry :pray:

danimontoya: Sure, no problem!! I'll change it back for `orEmpty` api only.. but I have a question about that too, for some typeclasses like `MapK`, `SetK` and `SortedMapK` there wasn't the possibility to use `Applicative`. Those can be skipped.. or shall I use `Id`/`Option.applicative()` ?

1Jajen1: Those cannot use `orEmpty` anyway so that doesn't need to be tested in that case. They can use `traverse_` and `sequence_` because that works with any applicative (just use `Id` or `Option`) and thus it needs to be tested

danimontoya: Something like this ?

arrow-core-test/src/main/kotlin/arrow/test/laws/FoldableLaws.kt
```diff
@@ -2,41 +2,110 @@ package arrow.test.laws
...
 
+  private fun <F> laws2(FF: Foldable<F>, GA: Applicative<F>, GENK: GenK<F>, EQK: EqK<F>): List<Law> {
+    val GEN: Gen<Kind<F, Int>> = GENK.genK(Gen.intSmall())
+    val EQKUnit = EQK.liftEq(Eq.any())
+
+    return listOf(
+      Law("Foldable Laws: traverse_ consistent with foldRight") { FF.`traverse_ consistent with foldRight`(GA, GENK, GEN, EQKUnit) },
+      Law("Foldable Laws: sequence_ consistent with traverse_") { FF.`sequence_ consistent with traverse_`(GA, GENK, EQKUnit) }
+    )
+  }
```
1Jajen1: Just integrate this in the main laws that don't use applicative instance for `F`. For the applicative instance just use `Id`.

arrow-core-test/src/main/kotlin/arrow/test/laws/FoldableLaws.kt
1Jajen1: Almost, no need to test `traverse_` and `sequence_` separately. But this is a good start.

arrow-core-test/src/main/kotlin/arrow/test/laws/FoldableLaws.kt
```diff
@@ -1,42 +1,128 @@
...
+    val GENB = GENK.genK(Gen.bool())
 
-    return listOf(
+    val EQ: Eq<Int> = Int.eq()
+    val EQBool = Boolean.eq()
+    val EQOptionInt = Option.eq(Int.eq())
+    val EQListKInt: ListKEq<Int> = ListK.eq(Int.eq())
+    val EQForIdInt: Eq<Kind<ForId, Int>> = Id.eqK().liftEq(Int.eq())
+
+    return laws2(Id.traverse(), Id.applicative(), Id.genK(), Id.eqK()) +
```
1Jajen1: This is pointless. `laws2` does not need to have `Foldable` and `Applicative` for `F`. `traverse_` and `sequence_` work with any applicative. So use `Foldable<F>` (`FF`) and any applicative for example `Id`. Also please inline `laws2`, it does not need to be seperate.

arrow-core-test/src/main/kotlin/arrow/test/laws/FoldableLaws.kt
```diff
@@ -1,42 +1,128 @@
...
+        Law("Foldable Laws: isNotEmpty consistent with isEmpty") { FF.`isNotEmpty consistent with isEmpty`(GEN, EQBool) },
+        Law("Foldable Laws: foldMapM folds on F mapping values to G(B) using given Monoid") { FF.`foldMapM folds on F mapping values to G(B) using given Monoid`(GEN, EQForIdInt) },
+        Law("Foldable Laws: get gets the item at the given index of the Foldable") { FF.`get gets the item at the given index of the Foldable`(GEN, EQOptionInt) }
+      )
+  }
+
+  fun <F> laws(FF: Foldable<F>, GA: Applicative<F>, EQK: EqK<F>): List<Law> {
+    val EQKListKInt: Eq<Kind<F, Kind<ForListK, Int>>> = EQK.liftEq(ListK.eq(Int.eq()))
+
+    return laws2(Option.traverseFilter(), Option.applicative(), Option.genK(), Option.eqK()) +
```
1Jajen1: This should not be needed. The main laws should already include `traverse_` and `sequence_`.

arrow-core-test/src/main/kotlin/arrow/test/laws/FoldableLaws.kt
danimontoya: Hi @1Jajen1 I will need some help to do that because I don't see the way to do it. If I use an instance of applicative Id.. I will have to change the generated function and Eq to check `ForId` and not any `F`. Thats why they were not inline within all laws together. Any hint is welcome!

1Jajen1: `fun <F, G> Foldable<F>.traverse_Test(AP: Applicative<G>, GF: GenK<F>, GG: GenK<G>, EQF: EqK<F>, EQG: EqK<G>)` and then basically the test you added below with data generated by these generators and validated by the eq instances.
In the laws just invoke them with any applicative of your choice. `traverseTest(Option.applicative(), GENK, Option.genK(), EQK, Option.eqK())`.

Basically just get rid of the constraint that for invoking `traverse_` you need the applicative and foldable for the **same** `F`. That is simply not the case.

danimontoya: Thanks @1Jajen1 , will try this 👍

danimontoya: Made the change, how does it look like now?

arrow-core-test/src/main/kotlin/arrow/test/laws/FoldableLaws.kt
```diff
@@ -1,42 +1,114 @@
...
-        Law("Foldable Laws: Exists is lazy") { FF.existsIsLazy(GEN, EQ) },
-        Law("Foldable Laws: ForAll is lazy") { FF.forAllIsLazy(GEN, EQ) },
-        Law("Foldable Laws: ForAll consistent with exists") { FF.forallConsistentWithExists(GEN) },
-        Law("Foldable Laws: ForAll returns true if isEmpty") { FF.forallReturnsTrueIfEmpty(GEN) },
-        Law("Foldable Laws: FirstOption returns None if isEmpty") { FF.firstOptionReturnsNoneIfEmpty(GEN) },
-        Law("Foldable Laws: FirstOption returns None if predicate fails") { FF.firstOptionReturnsNoneIfPredicateFails(GEN) },
-        Law("Foldable Laws: FoldM for Id is equivalent to fold left") { FF.foldMIdIsFoldL(GEN, EQ) }
-      )
+      Law("Foldable Laws: orEmpty consistent with just empty") { FF.`orEmpty consistent with just empty`(GA, EQKListKInt) }
+    )
```
1Jajen1: This never tests any of the other laws right? Or am I missing something here?

arrow-core-test/src/main/kotlin/arrow/test/laws/FoldableLaws.kt
```diff
@@ -1,42 +1,114 @@
...
 
+  fun <F> Foldable<F>.`foldRight is lazy`(G: Gen<Kind<F, Int>>, EQ: Eq<Int>) =
+    forAll(G) { fa: Kind<F, Int> ->
+      val sideEffect = SideEffect()
+      fa.foldRight(Eval.now(0)) { _, _ ->
+        sideEffect.increment()
+        Eval.now(1)
+      }.value()
+      val expected = if (fa.isEmpty()) 0 else 1
+      sideEffect.counter.equalUnderTheLaw(expected, EQ)
```
1Jajen1: I am not quite sure what exactly this is supposed to test. It executes the `Eval` returned by `foldRight` directly so it is not lazy at all. You should assign the eval to a value, check if the counter is still at 0 and then call eval and verify it being at 1.

1Jajen1: Also `equalUnderTheLaw` is a bit too much when comparing integers, just use the standart matchers like `should` directly.

arrow-core-test/src/main/kotlin/arrow/test/laws/FoldableLaws.kt
```diff
@@ -114,4 +181,145 @@ object FoldableLaws {
...
+    }
+
+  fun <F> Foldable<F>.`orEmpty consistent with just empty`(GA: Applicative<F>, EQ: Eq<Kind<F, Kind<ForListK, Int>>>) =
+    GA.run {
+      with(ListK.monoid<Int>()) {
+        orEmpty(this@run, this).equalUnderTheLaw(just(this.empty()), EQ)
+      }
+    }
+
+  fun <F, G> Foldable<F>.`traverse consistent with foldRight`(GA: Applicative<G>, GF: Gen<Kind<F, Int>>, GG: GenK<G>, EQG: Eq<Kind<G, Unit>>) =
```
1Jajen1: Rename to `traverse_` is probably better as `traverse` is quite different.

arrow-core-test/src/main/kotlin/arrow/test/laws/FoldableLaws.kt
```diff
@@ -114,4 +181,145 @@ object FoldableLaws {
...
+    forAll(Gen.functionAToB<Int, Kind<G, Int>>(GG.genK(Gen.intSmall())), GF) { f: (Int) -> Kind<G, Int>, fa: Kind<F, Int> ->
+      GA.run {
+        val expected = fa.foldRight(always { GA.just(Unit) }) { a, acc -> GA.run { f(a).apEval(acc.map { it.map { { _: Int -> Unit } } }) } }.value()
+        fa.traverse_(this, f).equalUnderTheLaw(expected, EQG)
+      }
+    }
+
+  fun <F, G> Foldable<F>.`sequence consistent with traverse`(GA: Applicative<G>, GF: GenK<F>, GG: GenK<G>, EQ: Eq<Kind<G, Unit>>) =
```
1Jajen1: Same as with the above law. Rename to `sequence_` and `traverse_`

arrow-core-test/src/main/kotlin/arrow/test/laws/FoldableLaws.kt
```diff
@@ -1,42 +1,114 @@
...
+      Law("Foldable Laws: combineAll is an alias for fold") { FF.`combineAll consistent with fold (alias)`(GENListK, EQListKInt) },
+      Law("Foldable Laws: reduceLeftToOption combines all items into an optional value") { FF.`reduceLeftToOption returns Option value`(GENB, EQOptionInt) },
+      Law("Foldable Laws: reduceRightToOption combines all items into an optional value") { FF.`reduceRightToOption returns Option value`(GENB, EQOptionInt) },
+      Law("Foldable Laws: reduceLeftOption consistent with reduceLeftToOption") { FF.`reduceLeftOption returns Option value`(GEN, EQOptionInt) },
+      Law("Foldable Laws: reduceRightOption consistent with reduceRightToOption") { FF.`reduceRightOption returns Option value`(GEN, EQOptionInt) },
+      Law("Foldable Laws: traverse_ consistent with foldRight") { FF.`traverse consistent with foldRight`(Id.applicative(), GEN, Id.genK(), Id.eqK().liftEq(Eq.any())) },
+      Law("Foldable Laws: sequence_ consistent with traverse_") { FF.`sequence consistent with traverse`(Id.applicative(), GENK, Id.genK(), Id.eqK().liftEq(Eq.any())) },
```
1Jajen1: Not entirely sure here, so do whatever you think is best:
Since we have only `Kind<F, Unit>` as a return value to compare we are not really checking all that much with `Id`. It's probably better to use a `F` that actually has some context like `Option`, `Either`, `Ior` etc.

arrow-core-test/src/main/kotlin/arrow/test/laws/FoldableLaws.kt
danimontoya: Mmm aren't the other laws tested through a different path? or I should call them from here too??

1Jajen1: Well if they are fine, but please test that because it does not look like it to me. Iirc in most other tests where we have additional tests for additional instances we call them in this place.

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -722,7 +722,7 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
   fun <C> foldRight(initial: Eval<C>, rightOperation: (B, Eval<C>) -> Eval<C>): Eval<C> =
     fix().let { either ->
       when (either) {
-        is Right -> rightOperation(either.b, initial)
+        is Right -> Eval.defer { rightOperation(either.b, initial) }
```
aballano: 👍

arrow-core-data/src/main/kotlin/arrow/typeclasses/Foldable.kt
```diff
@@ -241,7 +241,11 @@ interface Foldable<F> {
   fun <A> Kind<F, A>.firstOrNone(predicate: (A) -> Boolean): Option<A> =
     find { predicate(it) }
 
-  fun <A> Kind<F, A>.toList(): List<A> = foldRight(Eval.now(emptyList<A>())) { v, acc -> acc.map { listOf(v) + it } }.value()
+  /**
+   * Turn items into a list
+   */
```
aballano: nit: I don't think the comment is adding much, so we could remove it, no?

```suggestion
```

1Jajen1: Agreed, it's also not really accurate. "Turning structure into list" would be better, but `toList` is descriptive enough imo.

PR #49, opened 2020-02-27
add typeclasses MonadPlus & MonadLogic

abendt:
add typeclasses MonadPlus & MonadLogic
with instances for ListK/Option/SequenceK

this continues https://github.com/arrow-kt/arrow/pull/2047
part of https://github.com/arrow-kt/arrow-incubator/issues/5


1Jajen1:
I have some time to review this tomorrow. I'll also try and see if I can get the sequence problem sorted out... If not we should document that so others don't fall over it.

It is however really important that at some point we have an implementation of `LogicT` in transformers. That is pretty hard to do, but it's the only good instance for `MonadLogic` especially if `Sequence` has these problems.

arrow-core-test/src/main/kotlin/arrow/test/laws/MonadLogicLaws.kt
```diff
@@ -0,0 +1,135 @@
...
+import io.kotlintest.properties.forAll
+import io.kotlintest.shouldBe
+
+object MonadLogicLaws {
+
+  fun <F> laws(
+    ML: MonadLogic<F>,
+    GK: GenK<F>,
+    EQK: EqK<F>,
+    iterations: Int = 1000
```
1Jajen1: What is the idea behind this? This is too low for stacksafety and too high for normal testing isn't it?

1Jajen1:
Do you want to add some more docs to this or experiment with it? Otherwise I'd be fine with merging it as is. (I'll probably play around with it a bit^^)

After merging it needs the instances for `mtl`. You might be able to do that with the previous branch as well.

Regarding the `LogicT` datatype. This will be hard to do. There are two ways to emulate rank-n types in kotlin (that I know of):
You can either do it explicitly using an interface:
```kotlin
interface T {
  operator fun <A> invoke(): A
}
```
Problem with this approach: It is really hard to make this stacksafe. Not sure it is worth the trouble.

Second approach: Poor mans rank-n type.
```kotlin
data class T(val x: Any?)
```
This usually leads to a ton of typecasts, but since those are likely 100% internal it may not be a problem.

For `LogicT` this means:
`data class LogicT<M, A>(val cont: ((A, Kind<M, Any?>) -> Kind<M, Any?>, Kind<M, Any?>) -> Kind<M, Any?>)`. Since all the `Any?` stuff would be internal it would not be a huge problem. This could then be exposed as `fun <M, A, R> LogicT<M, A>.runLogicT(f: (A, Kind<M, R>) -> Kind<M, R>, e: Kind<M, R>): Kind<M, R>`.

The interface solution would be:
```kotlin
interface LogicT<M, A> {
  fun <R> runLogicT(f: (A, Kind<M, R>) -> Kind<M, R>, e: Kind<M, R>): Kind<M, R>
}
```

I might give implementing it a shot, but if I do I'll give it a note in the issue.

arrow-core-test/src/main/kotlin/arrow/test/laws/MonadLogicLaws.kt
abendt: i had some issues with long running tests for StateT so i added an option to control the number of iterations. in my understanding 1000 here is the default used by kotlintest.

https://github.com/arrow-kt/arrow/pull/2047/files#diff-55e755bcfe1e88678c7c35d4aec3432bR93

tbh: i didn't understand what was the underlying reason for the slow tests ...

abendt:
> Do you want to add some more docs to this or experiment with it? Otherwise I'd be fine with merging it as is. (I'll probably play around with it a bit^^)
> 

at the time i don't have much time to spare. so if its ok just merge it. i might add some enhanced docs in the future

> After merging it needs the instances for `mtl`. You might be able to do that with the previous branch as well.

there are some mtl instances in the old branch: https://github.com/arrow-kt/arrow/pull/2047

i can create a new PR when this was merged

1Jajen1:
Then I'll leave the issue for it open and add some information for what is left to do.

1Jajen1:
> Then I'll leave the issue for it open and add some information for what is left to do.

Done. Thanks again for contributing :heart: @abendt 

It will be interesting to see where this can go... Logic programming is very interesting for more some algorithms.

PR #54, opened 2020-03-02
Fix dependencies

rachelcarmena:
Fix https://github.com/arrow-kt/arrow-optics/issues/24 and https://github.com/arrow-kt/arrow-incubator/issues/53 for Arrow Core


rachelcarmena:
Thanks mates!! :raised_hands:

PR #55, opened 2020-03-04
Fix missing 'Type Class Hierarchy' section in website

rachelcarmena:


rachelcarmena:
Thanks @aballano :raised_hands:

PR #56, opened 2020-03-08
Fix small bug in MonadLogic

1Jajen1:
This is a nasty little bug. Using `unit()` called `map { Unit }` which would mean it usually diverges (because pruning usually works on infinite values). Using `just(Unit)` also doesn't work because that infers to `A.just(dummy: Unit = Unit)`... 

This usually does not show in tests because most tests are for finite/small values.


arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadLogic.kt
```diff
@@ -157,7 +157,8 @@ interface MonadLogic<F> : MonadPlus<F> {
    * }
    */
   fun <A> Kind<F, A>.voidIfValue(): Kind<F, Unit> =
-    once().ifThen(unit()) { zeroM() }
+    once().ifThen(Unit.just()) { zeroM() }
+  // TODO fix this when the name ambiguities of unit() are removed and not just deprecated.
```
nomisRev: Ouch. Good things this will be gone in 0.11.0!

PR #64, opened 2020-03-10
Undo unnecessary change

rachelcarmena:
It's not necessary and it impacts on the use of the libraries (necessary dependencies).

I don't use "revert" to keep some improvements/orders done.


rachelcarmena:
Thanks for the review @aballano :raised_hands:

PR #66, opened 2020-03-11
Check binary compat diff 0.10.5 .api files

danimontoya:


nomisRev:
Thanks @danimontoya for looking into this.
As you mentioned on Slack we have a breaking change from `0.10.4` to `0.10.5` in `DoubleSemiring`.

This happened in https://github.com/arrow-kt/arrow/pull/2076.

> I also changed Semiring to not inherit from Monoid because that is ambiguos because a semiring is basically two monoids combined with a bunch of extra laws. I also added the law that the additive part has to be commutative. This ruled out the Schedule instance. I also removed the Option instance because it failed the laws as well.

The `Schedule` instance can safely be removed since it isn't released yet.

These are all breaking change:
 -  `Option` instance could simply be marked for removal with `@Deprecated`.
 - `Monoid` removal from `Semiring` however is a tricky breaking change, since deprecation will not help us in this case. Can we find another solution to fix this issue @1Jajen1? A next major release, or a migration to Arrow Meta would allow us to make this change easier.

1Jajen1:
I can't remember if this was breaking stuff or not. If not we an just readd it for this version. But yeah the next major one should definitely remove it.

danimontoya:
Thanks for the feedback, closing this since it was only to show the diff between 0.10.4 and 0.10.5

PR #70, opened 2020-03-16
Bump versions of KOTLIN_VERSION and KOTLIN_TEST_VERSION locally for core

danimontoya:
Bump versions of KOTLIN_VERSION=1.3.70 and KOTLIN_TEST_VERSION=3.4.2. All test passed 
[core-dependencies.txt](https://github.com/arrow-kt/arrow-core/files/4338038/core-dependencies.txt)

### DIFF
--- a/arrow-syntax/build.gradle
+++ b/arrow-syntax/build.gradle
@@ -14,7 +14,7 @@ apply from: "$DOC_CONF"
 apply from: "$PUBLISH_CONF"
 
 dependencies {
-    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$KOTLIN_VERSION"
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$KOTLIN_VERSION"
     compile project(":arrow-annotations")
     compile project(":arrow-core")
     testRuntime "org.junit.vintage:junit-vintage-engine:$JUNIT_VINTAGE_VERSION"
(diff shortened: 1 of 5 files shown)


gradle.properties
```diff
@@ -12,3 +12,5 @@ kotlin.incremental=true
 # Kotlin Test configuration
 #Parallelism needs to be set to 1 since the concurrent tests in arrow-effects become flaky otherwise
 kotlintest.parallelism=1
+KOTLIN_VERSION=1.3.70
+KOTLIN_TEST_VERSION=3.4.2
```
rachelcarmena: Please, don't add those versions here. They come from `arrow` repository.

arrow-syntax/build.gradle
```diff
@@ -14,7 +14,7 @@ apply from: "$DOC_CONF"
 apply from: "$PUBLISH_CONF"
 
 dependencies {
-    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$KOTLIN_VERSION"
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$KOTLIN_VERSION"
```
rachelcarmena: I think we shouldn't change it because of these comments: https://github.com/arrow-kt/arrow/pull/2065

danimontoya: Good one, thanks!

gradle.properties
danimontoya: Agree!! I'm closing this PR.. since bumping the versions worked and all test passed.. we can move on and update it in the general repo 👍

PR #71, opened 2020-03-17
Fix apply order of product to be consistent with other methods

1Jajen1:
Fixes a few ordering problems that I missed in earlier passes.


1Jajen1:
Thanks @pakoito 
At some point this needs better testing. All the product-*, tupledN, mapN methods that is, however I can't think of a good way to check ordering because making it part of Apply would be wrong as that alone does not imply any ordering. The fact that it is left-to-right in arrow is merely convention (a good one, but not a requirement)

PR #74, opened 2020-03-25
0.11.0-SNAPSHOT wip branch

rachelcarmena:
I'm creating this pull request because this branch should be merged as soon as `0.10.5` is released in order to allow the following contributions for `0.11.0-SNAPSHOT`.

Sync bot is checking build result. However, it's not checking API doc generation and validation because of posponed errors. Let's create this pull request to check the status with `0.11.0-SNAPSHOT`.

**Note**: this pull request should change some files before merging like `gradle.properties`, sync bot, etc. as soon as `0.10.5` is released (because `VERSION_NAME` will be `0.11.0-SNAPSHOT` then). They are included in #76


arrow-core-data/src/main/kotlin/arrow/typeclasses/suspended/BindSyntax.kt
```diff
@@ -23,4 +24,9 @@ interface BindSyntax<F> {
 
   suspend operator fun <A> Kind<F, A>.not(): A =
     bind()
+
+  suspend fun <A> Kind<F, A>.getValue(thisRef: Any?, property: KProperty<*>): A =
+    bind()
 }
+
+operator fun String.getValue(thisRef: Any?, property: KProperty<*>): Char = 'c'
```
nomisRev: Why where these  2 `getValue` operators added?

raulraja: I believe the above for the bind syntax to see if we can enable it already to start using it but the String.getValue makes no sense

rachelcarmena: Simon, I think they come from `0.11.0/bio` branch: https://github.com/arrow-kt/arrow-core/pull/74/commits/99bceecd15976cc390ab0eaeb91ee95ad0853bc7

PR #77, opened 2020-04-01
Rename artifacts to fix inconsistency

lucmazon:
Relates to https://github.com/arrow-kt/arrow-integrations/issues/19


rachelcarmena:
It's already available in the next version: https://arrow-kt.io/docs/next/quickstart/libraries/index.html

It seems it's also necessary to fix the "table of content". I'm going to prepare that PR :raised_hands:

PR #86, opened 2020-04-03
Try to fix ListKCrosswalk instance with Proguard

### LINKED ISSUE #85: ListK Crosswalk instance fails the build with Proguard enabled (Android)
The Crosswalk instance is giving this error when proguard is enabled:

> Warning: arrow.core.extensions.ListKCrosswalk$crosswalk$1$2$1$1: can't find referenced class arrow.core.extensions.ListKCrosswalk$crosswalk$1$2

More info in https://github.com/arrow-kt/arrow/issues/807
Related to https://youtrack.jetbrains.com/issue/KT-16084

aballano:
Related to https://youtrack.jetbrains.com/issue/KT-16084

Potentially fixes #85


aballano:
@pakoito @rachelcarmena is it possible to test this on an android project before merging? Would need to verify that it works with Proguard

rachelcarmena:
> @pakoito @rachelcarmena is it possible to test this on an android project before merging? Would need to verify that it works with Proguard

Good idea!! It would be great if we have a project to test it and we could include its verification during the build check :+1:

nomisRev:
@rachelcarmena @aballano only thing I can imagine is some kind of Android/Arrow build we compile as part of a community build.

aballano:
@nomisRev @rachelcarmena What do you think folks about merging this PR and wait for https://github.com/arrow-kt/arrow-fx/pull/158 to then create a separate repo with an example android app where we add some examples to verify this kind of issues?

rachelcarmena:
> @nomisRev @rachelcarmena What do you think folks about merging this PR and wait for [arrow-kt/arrow-fx#158](https://github.com/arrow-kt/arrow-fx/pull/158) to then create a separate repo with an example android app where we add some examples to verify this kind of issues?
> 
> My current problem here is that I'd need to add an extra module only to test this, and then we probably are gonna need another one for Fx, so maybe extracting an example android module outside we can merge together all tests/checks in a single place.

Sounds good to me :+1: :raised_hands:

PR #90, opened 2020-04-09
Clean: remove unnecessary function

rachelcarmena:
It comes from https://github.com/arrow-kt/arrow-core/pull/74#pullrequestreview-383721919


rachelcarmena:
Thanks for the review @aballano :raised_hands:

PR #94, opened 2020-04-15
Revamp Kind encodings

### LINKED ISSUE #88: Revamp Kind encoding to support N-Functor Hierarchies
https://github.com/arrow-kt/arrow-core/commit/b6710925e02b844c6df47bc9959a0be1d1df758e

The new Kind encoding should change to flatten partial application instead of using nesting over `F`.

This allows for:
```
`Kind<F, A>` =!= `Kind22<F, ..Nothing..., A>` to support `Kind2<F, E, A>` that can use `Functor.map` and other methods designed for `Kind<F, A>`
```
This also allows us to establish a proper effects hierarchy and get rid of `MonadThrow` bringing `raiseError` for `Throwable` up to `MonadDefer` or wherever necessary if we consider instantiating `Throwable` an effect.

This also opens the door to support Bifunctor behaviors for Rx, Reactor, and others with specialized data types that bake in EitherT like functionality.

This refactor is for 0.11 so we can follow the following steps:

- Release 0.11.0 with Bifunctor IO and no changes to the kind hierarchy
- Release 0.12.0 with Bifunctor type classes and potentially bi-functor Apis for Rx and Reactor

This change impacts meta IDE features related to folding kinds and current proofs for kinds in meta and it's part of the first migration of the meta prelude to arrow-core to flatten the encoding.

aballano:
Some considerations:

* Due to [this change](https://github.com/arrow-kt/arrow-core/pull/94/files#diff-22d8a4f289b2a7fbff40cdd08ed30220R206), the return types in the generated instances will appear in their Kinded form, instead of the typealiased one.
* @raulraja I finally left the partial application of Kinds as you did (`Kind2<out F, out A, out B> : Kind3<F, A, Nothing, B>`), although we might want to test with an alternative left-encoding (`Kind2<out F, out A, out B> : Kind3<F, Nothing, A, B>`) if it makes sense whatsoever.

Fixes #88


arrow-core/src/main/kotlin/arrow/core/extensions/listk.kt
```diff
@@ -130,12 +129,6 @@ interface ListKMonad : Monad<ForListK> {
   override fun <A, B> Kind<ForListK, A>.map(f: (A) -> B): ListK<B> =
     fix().map(f)
 
-  // Special case, since the extension is built with List<List<A>>, then wrapped to ListK<List<A>>, there's no way for that to
-  // call the default flatten which is defined for ListK<ListK<A>> due to the inner list, therefore we need to map and redirect.
-  fun <A> Kind<ForListK, List<A>>.flatten(dummy: Unit = Unit): Kind<ForListK, A> =
-    // explicit cast for proper resolution as ListK extends List
-    (map { it.k() } as Kind<ForListK, Kind<ForListK, A>>).flatten()
-
```
aballano, line 138: Some goodies due to the change 😍

arrow-core-data/src/main/kotlin/arrow/typeclasses/Cocomposed.kt
```diff
@@ -1,40 +0,0 @@
-package arrow.typeclasses
```
pakoito, line 1: RIP in Pieces

arrow-core/src/main/kotlin/arrow/core/extensions/andthen.kt
```diff
@@ -44,45 +40,45 @@ interface AndThenMonoid<A, B> : Monoid<AndThen<A, B>>, AndThenSemigroup<A, B> {
 }
 
 @extension
-interface AndThenFunctor<X> : Functor<AndThenPartialOf<X>> {
-  override fun <A, B> AndThenOf<X, A>.map(f: (A) -> B): AndThen<X, B> =
-    fix().map(f)
+interface AndThenFunctor : Functor<ForAndThen> {
+  override fun <A, B> AndThenPartialOf<A>.map(f: (A) -> B): AndThenPartialOf<B> =
```
pakoito, line 44: I believe this may not be correct. We want to return concrete instances from the functions, so `AndThen<X, B>` is what's expected as a return.

pakoito: This applies to the others too. I ned to check the rest of the diff.

aballano: Good point, i'll take another look then, thanks mate ❤️

aballano:
Closing for now as we will follow a different path here 😄

PR #96, opened 2020-04-17
Fix error handling examples

### LINKED ISSUE #95: Error handling code samples don't compile
Embarrassingly, the [`Option` code samples](https://arrow-kt.io/docs/0.10/patterns/error_handling/#option) don't compile.

The import of `arrow.core.extensions.option.monad.binding` can't be resolved. Apparently `fx` is in there, because it can't be resolved. The same goes for `bind()`. IntelliJ doesn't come up with suggestions on where to import `fx` from.

rachelcarmena:
Resolves #95 thanks to @JorgeCastilloPrz


aballano:
Thanks @rachelcarmena!

PR #99, opened 2020-04-24
Fix broken links

rachelcarmena:


rachelcarmena:
Thanks for the review @calvellido :raised_hands:

PR #100, opened 2020-04-26
Fixing broken link in the introduction

sierisimo:
The current documentation sends to a wrong link for the implementation of `Option`. 

Simply removing the part of the URL should fix it.


pakoito:
Good catch. We may have a few more of these left over!

rachelcarmena:
Thanks @sierisimo :raised_hands: I've run `wget` with `spider` option to catch more broken links: #101

PR #102, opened 2020-04-30
Adding conditionally as a fix for order of cond

### LINKED ISSUE #98: Either.cond parameter order mismatch
All of our APIs have errors on the left...except for that one.

https://github.com/arrow-kt/arrow-core/blob/master/arrow-core-data/src/main/kotlin/arrow/core/Either.kt#L866

As seen in https://youtu.be/tN_k1W1hEhA?t=7094

sierisimo:
This commit resolves #98 with the new function `conditionally` as
proposed in Slack by @hiosdra

It also moves `cond` to just pass it's parameters to `conditionally`

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
@@ -416,7 +416,9 @@ import arrow.typeclasses.Show
  * }
  * ```
  *
- * For creating Either instance based on a predicate, use `Either.conditionally()` method :
+ * For creating Either instance based on a predicate, use `Either.conditionally()` method. It will evaluate an expression
+ * passed as first parameter, in case the expression evaluates to `false` it will give an `Either.Left<L>` build from the second parameter.
+ * If the expression evaluates to a `true` it will take the third parameter and give an `Either.Right<R>`:
  *
  * ```kotlin:ank:playground
  * import arrow.core.Either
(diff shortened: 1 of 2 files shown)


Hiosdra:
Great 👍 
One thing ->  Shoudn't `Either.cond` be made deprecated in favour of new method if we move to `conditionally` in tests?

sierisimo:
I wanted baby steps on moving out `cond` but I can add the `@Deprecated(…)`

Or should we just kill it with fire? 

WITH FIRE I SAID

pakoito:
Let's be nice and deprecate first :D

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -416,14 +416,14 @@ import arrow.typeclasses.Show
  * }
  * ```
  *
- * For creating Either instance based on a predicate, use `Either.cond()` method :
+ * For creating Either instance based on a predicate, use `Either.conditionally()` method :
```
pakoito, line 419: could you please add something about parameter order here?

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -863,7 +863,9 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
       }
     }
 
-    fun <L, R> cond(test: Boolean, ifTrue: () -> R, ifFalse: () -> L): Either<L, R> = if (test) right(ifTrue()) else left(ifFalse())
+    fun <L, R> cond(test: Boolean, ifTrue: () -> R, ifFalse: () -> L): Either<L, R> = conditionally(test, ifFalse, ifTrue)
+
+    fun <L, R> conditionally(test: Boolean, ifFalse: () -> L, ifTrue: () -> R): Either<L, R> = if (test) right(ifTrue()) else left(ifFalse())
```
pakoito, line 868: Some docs atop this would be nice. Not a requierement for merging, just nice.

sierisimo:
Docs added to explain how this works!

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -863,7 +865,28 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
       }
     }
 
-    fun <L, R> cond(test: Boolean, ifTrue: () -> R, ifFalse: () -> L): Either<L, R> = if (test) right(ifTrue()) else left(ifFalse())
+    @Deprecated(
+      message = "use conditionally as parameters are consistent with Either class",
```
pakoito, line 869: *parameter order

pakoito:
To prod!

PR #103, opened 2020-05-05
 Implementation of Can<A, B>

pablisco:
Implementation of `Can`from Haskell: https://hackage.haskell.org/package/smash-0.1.1.0/docs/Data-Can.html#v:fold

_Related to https://github.com/arrow-kt/arrow-incubator/issues/108_


arrow-core-data/src/test/kotlin/arrow/core/CanTest.kt
```diff
@@ -0,0 +1,415 @@
...
+        { (a to b).toCan() to Can.Both(a, b) },
+        { (a toT b).toCan() to Can.Both(a, b) },
+        { Ior.Left(a).toCan() to Can.Left(a) },
+        { Ior.Right(b).toCan() to Can.Right(b) },
+        { Ior.Both(a, b).toCan() to Can.Both(a, b) },
+        { None.toRightCan() to Can.Neither },
+        { Some(b).toRightCan() to Can.Right(b) },
+        { None.toLeftCan() to Can.Neither },
+        { Some(a).toLeftCan() to Can.Left(a) }
+      )
```
pablisco, line 98: I'm not 100% sold on how I've implemented multi-case tests on this suite as they are hard to pin point failures when they happen since the whole groups fails. The idea of `allEqual` was to get better info, but doesn't work. 
I'm open to suggestions, although I may look at them on a follow-up PR.

raulraja:
Looks great but this needs docs and some kind of use case tutorial. The tests don't help understand how it's useful.

pablisco:
Added extra documentation :)

aballano:
As per [this discussion](https://github.com/arrow-kt/arrow-core/pull/111#issuecomment-631339275), this should be moved to the incubator first

pablisco:
Moved to https://github.com/arrow-kt/arrow-incubator/pull/109

PR #104, opened 2020-05-07
Update internal atomic wrapper

aballano:
Add missing getAndUpdate fun
Add missing return types
Add couple of inlines to high-order functions
Put declarations on new line


aballano:
@rachelcarmena as soon as this gets green I'd like to push a change for a more restrictive gradle config in which we only put the atomicfu deps on the module using it, if it makes sense 😅

rachelcarmena:
> @rachelcarmena as soon as this gets green I'd like to push a change for a more restrictive gradle config in which we only put the atomicfu deps on the module using it, if it makes sense sweat_smile

Of course! Thanks @aballano :raised_hands: Right, it's being used for the whole `arrow-core` and `arrow-fx` so far. Good idea!! :+1:

aballano:
> @rachelcarmena as soon as this gets green I'd like to push a change for a more restrictive gradle config in which we only put the atomicfu deps on the module using it, if it makes sense 😅

NVM it seems some atomic is used in core-data so we cannot restrict it too much for now

rachelcarmena:
> > @rachelcarmena as soon as this gets green I'd like to push a change for a more restrictive gradle config in which we only put the atomicfu deps on the module using it, if it makes sense sweat_smile
> 
> NVM it seems some atomic is used in core-data so we cannot restrict it too much for now

Or maybe it could be added just for both :raised_hands:

PR #105, opened 2020-05-08
[Discussion] UnitSpec simplification

suhtai:
This is a potential improvement for how to use `UnitSpec`.

Instead of relying on using a `unit {}` block this new implementation accepts two parameters, one for laws and a second one for the actual tests.

This also means that the test cases are created lazily when the method `testCases()` is invoked instead of at the time of constructing the spec.

*Question:* If we decide to go with this new pattern, should we change all the tests (should be straight forward) or leave `testLaws()` as deprecated and slowly evolve the code?

*Question:* Should we use `distinct { name }` to remove duplicate tests or should we instead let the tests fail? We are currently doing that with the laws. We could potentially be ignoring tests cases that are duplicated without our knowledge.


nomisRev:
> Question: Should we use distinct { name } to remove duplicate tests or should we instead let the tests fail? We are currently doing that with the laws. We could potentially be ignoring tests cases that are duplicated without our knowledge.

That shouldn't be the case, until recently at least a duplicated test name resulted in a crash. We were only `distinct`ing the laws. If this has changed, then we should change it back so duplicated concrete test names fail the test suite.

nomisRev:
`KotlinTest` now `Kotest` actually favors, or at least that's the one I have been using outside of Arrow. This way you can enforce "functions" as tests since you cannot rely on pre-after conditions etc.

```
class ExampleTest : UnitSpec({
   // ... 
})
```

nomisRev:
`KotlinTest` now `Kotest` actually favors, or at least that's the one I have been using outside of Arrow. This way you can enforce "functions" as tests since you cannot rely on pre-after conditions etc.

```
class ExampleTest : UnitSpec({
   // ... 
})
```

Ideally we could implement the laws unrelated to KotlinTest, but not sure which approach we can take there.

rachelcarmena:
Hi fellow contributor :wave: 

Arrow repositories has been reunified again into [`arrow`](https://github.com/arrow-kt/arrow) so this repository is going to be archived.

Please, if you want to replay this pull request on `arrow` repository, follow these steps:

```
# Adapt your branch to be able to be merged in the new arrow repository
cd arrow-core
git-filter-repo --to-subdirectory-filter arrow-libs/core
cd ..

# Create a new branch in arrow repository and merge your content
git clone git@github.com:arrow-kt/arrow.git
cd arrow
git checkout -b <new-branch>
git remote add arrow-core ../arrow-core
git fetch arrow-core
git merge --no-edit --allow-unrelated-histories arrow-core/<your-branch>
# check and create the pull request again on arrow repository
```

Please, let us know if we can help you!

Thank you so much for your contributions!! :raised_hands:

PR #108, opened 2020-05-15
make NonEmptyList Iterable

zraktvor:
# Change
Make NonEmptyList\<A\> Iterable\<A\> by delegation of NonEmptyList.all

# Reason
As of now the usage of NonEmptyList in functions like forEach or joinToString needs the indirection of calling NonEmptyList.all. This can be safed for cleaner code by making NonEmptyList Iterable. This also gives less unnecessary compiler errors when using NonEmptyList as a drop-in replacement for List from kotlin-stdlib. This should break no existing code.


aballano:
Thank you for the contribution @zraktvor!!

PR #111, opened 2020-05-20
[WIP]Add Smash as Option2

pablisco:
# ⚠️ Work In Progress ⚠️ 

Haskell has https://hackage.haskell.org/package/smash-0.1.1.0/docs/Data-Smash.html

Scalaz has Maybe2: https://scalaz.github.io/8/data/Maybe2.html

So to mix the best of both worlds we get `Option2` :) 

## What needs doing?

Moar tests and adding extensions for more type classes

_Related to https://github.com/arrow-kt/arrow-incubator/issues/108_


raulraja:
Hi @pablisco `Can`, `Smash` and `others` should be moved to the incubator as we did with other data types that while they may be legit and useful can't be in core because we are about to release 0.11 and freeze the repository for 1.0. 

Back then it was decided that datatypes without clear profit in Kotlin over what the language provides should not be in core. Would you mind relocating these PRs to the incubator in whatever module they belong to? 

Once proven useful we can promote them back to core or relevant maintained modules where needed. Thanks!

pablisco:
Moved to https://github.com/arrow-kt/arrow-incubator/pull/110

PR #112, opened 2020-05-20
Suspend Either.fx implementation

nomisRev:
One of the biggest issue we currently have in Arrow Core is when you try to use the data-types and APIs in combination with `suspend fun`.

The current `fx` implementation doesn't allow calling foreign `suspend` functions inside an `fx` block. This implementation solves that problem by introducing a state machine that keeps track of the `suspension` state.

This implementation can be copied to all other data types, and can thus offer an alternative to our currently non-suspending `fx` implementation.

With the `suspend` implementation it'll be possible to do the following.

```kotlin
import arrow.core.Either
import arrow.core.extensions.fx

data class User(val name: String)
object DomainError

suspend fun fetchDataFromNetwork(id: Int): Either<DomainError, User> = TODO()

suspend fun getUser(id: Int) : Either<DomainError, User> =
  Either.fx<DomainError, User> {
    // Restricted suspending functions can only invoke member or extension suspending functions on their restricted coroutine scope
    !fetchDataFromNetwork(id) 
  }
```

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
@@ -5,6 +5,10 @@ import arrow.core.Either.Left
 import arrow.core.Either.Right
 import arrow.higherkind
 import arrow.typeclasses.Show
+import kotlin.coroutines.Continuation
+import kotlin.coroutines.intrinsics.COROUTINE_SUSPENDED
+import kotlin.coroutines.intrinsics.startCoroutineUninterceptedOrReturn
+import kotlin.coroutines.intrinsics.suspendCoroutineUninterceptedOrReturn
 
 /**
  *
@@ -1052,11 +1056,11 @@ fun <A, B, C> EitherOf<A, B>.ap(ff: EitherOf<A, (B) -> C>): Either<A, C> =
 
 fun <A, B> EitherOf<A, B>.combineK(y: EitherOf<A, B>): Either<A, B> =
   when (this) {
-    is Either.Left -> y.fix()
+    is Left -> y.fix()
     else -> fix()
   }
 
-fun <A> A.left(): Either<A, Nothing> = Either.Left(this)
+fun <A> A.left(): Either<A, Nothing> = Left(this)
 
 fun <A> A.right(): Either<Nothing, A> = Either.Right(this)
 
@@ -1071,7 +1075,7 @@ fun <A> A.right(): Either<Nothing, A> = Either.Right(this)
  * ```
  */
 fun <A, B> B?.rightIfNotNull(default: () -> A): Either<A, B> = when (this) {
-  null -> Either.Left(default())
+  null -> Left(default())
   else -> Either.Right(this)
 }
 
@@ -1093,7 +1097,34 @@ fun <A> Any?.rightIfNull(default: () -> A): Either<A, Nothing?> = when (this) {
 fun <A, B> EitherOf<A, B>.handleErrorWith(f: (A) -> EitherOf<A, B>): Either<A, B> =
   fix().let {
     when (it) {
-      is Either.Left -> f(it.a).fix()
+      is Left -> f(it.a).fix()
       is Either.Right -> it
     }
   }
+
+suspend fun <E, A> Either.Companion.fx(c: suspend EitherContinuation<E, A>.() -> A): Either<E, A> =
+  suspendCoroutineUninterceptedOrReturn sc@{ cont ->
+    val continuation = EitherContinuation(cont as Continuation<EitherOf<E, A>>)
+    val wrapReturn: suspend EitherContinuation<E, A>.() -> Either<E, A> = { c().right() }
+
+    // Returns either `Either<A, B>` or `COROUTINE_SUSPENDED`
+    val x = try {
+      wrapReturn.startCoroutineUninterceptedOrReturn(continuation, continuation)
+    } catch (e: Throwable) {
+      if (e is SuspendMonadContinuation.ShortCircuit) Either.Left(e.e as E)
+      else throw e
+    }
+
+    return@sc if (x == COROUTINE_SUSPENDED) continuation.getResult()
+    else x as Either<E, A>
+  }
+
+class EitherContinuation<E, A>(
+  parent: Continuation<EitherOf<E, A>>
+) : SuspendMonadContinuation<EitherPartialOf<E>, A>(parent) {
+  override fun <A> Kind<EitherPartialOf<E>, A>.not(): A =
+    fix().fold({ e -> throw ShortCircuit(e) }, ::identity)
+
+  override fun ShortCircuit.recover(): Kind<EitherPartialOf<E>, A> =
+    Left(e as E)
+}
--- a/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
@@ -1052,7 +1052,7 @@ fun <A, B> EitherOf<A, B>.contains(elem: B): Boolean =
   fix().fold({ false }, { it == elem })
 
 fun <A, B, C> EitherOf<A, B>.ap(ff: EitherOf<A, (B) -> C>): Either<A, C> =
-  flatMap { a -> ff.fix().map { f -> f(a) } }.fix()
+  flatMap { a -> ff.fix().map { f -> f(a) } }
 
 fun <A, B> EitherOf<A, B>.combineK(y: EitherOf<A, B>): Either<A, B> =
   when (this) {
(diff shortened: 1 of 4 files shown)


pakoito:
LGTM. It puts a requirement that `fx` has to be suspended, and we have many cases where it's being used for pure data transformations. We should keep both IMO.

nomisRev:
@pakoito yes, that was the plan to keep both :) A non-suspend capable and a suspend capable version.

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -1093,7 +1097,34 @@ fun <A> Any?.rightIfNull(default: () -> A): Either<A, Nothing?> = when (this) {
 fun <A, B> EitherOf<A, B>.handleErrorWith(f: (A) -> EitherOf<A, B>): Either<A, B> =
   fix().let {
     when (it) {
-      is Either.Left -> f(it.a).fix()
+      is Left -> f(it.a).fix()
       is Either.Right -> it
```
pakoito, line 1101: Worth fixing both here, I guess :P

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -1093,7 +1097,34 @@ fun <A> Any?.rightIfNull(default: () -> A): Either<A, Nothing?> = when (this) {
...
     }
   }
+
+suspend fun <E, A> Either.Companion.fx(c: suspend EitherContinuation<E, A>.() -> A): Either<E, A> =
+  suspendCoroutineUninterceptedOrReturn sc@{ cont ->
+    val continuation = EitherContinuation(cont as Continuation<EitherOf<E, A>>)
+    val wrapReturn: suspend EitherContinuation<E, A>.() -> Either<E, A> = { c().right() }
+
+    // Returns either `Either<A, B>` or `COROUTINE_SUSPENDED`
```
pakoito, line 1110: Add the `Any` explicitly then.

arrow-core-data/src/main/kotlin/arrow/core/SuspendingMonadContinuation.kt
```diff
@@ -0,0 +1,81 @@
...
+    _decision.loop { decision ->
+      when (decision) {
+        UNDECIDED -> {
+          val r: Kind<F, A>? = when {
+            result.isFailure -> {
+              val e = result.exceptionOrNull()
+              if (e is ShortCircuit) e.recover() else null
+            }
+            result.isSuccess -> result.getOrNull()
+            else -> throw RuntimeException("IMPOSSIBLE")
```
pakoito, line 47: NIT: Unique exception here, cached in an object

arrow-core-data/src/test/kotlin/arrow/core/EitherTest.kt
```diff
@@ -203,5 +210,83 @@ class EitherTest : UnitSpec() {
...
+      Gen.bind(Gen.int(), Gen.throwable(), ::Pair)
+        .random()
+        .take(1001)
+        .forEach { (i, exception) ->
+          shouldThrow<Throwable> {
+            Either.fx<String, Int> {
+              val res = !Either.Right(i)
+              throw exception
+              res
+            } shouldBe Either.Right(i)
```
pakoito, line 252: This assertion should never happen. Use `fail` instead.

pakoito: There's another one on the next test.

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -1093,7 +1097,34 @@ fun <A> Any?.rightIfNull(default: () -> A): Either<A, Nothing?> = when (this) {
...
+    }
+
+    return@sc if (x == COROUTINE_SUSPENDED) continuation.getResult()
+    else x as Either<E, A>
+  }
+
+class EitherContinuation<E, A>(
+  parent: Continuation<EitherOf<E, A>>
+) : SuspendMonadContinuation<EitherPartialOf<E>, A>(parent) {
+  override fun <A> Kind<EitherPartialOf<E>, A>.not(): A =
```
pakoito, line 1125: We're moving away from `not` syntax, use `bind` or `BindSyntax` instead.

arrow-core-data/src/main/kotlin/arrow/core/SuspendingMonadContinuation.kt
```diff
@@ -0,0 +1,81 @@
...
+  override val context: CoroutineContext = EmptyCoroutineContext
+
+  override fun resumeWith(result: Result<Kind<F, A>>) {
+    _decision.loop { decision ->
+      when (decision) {
+        UNDECIDED -> {
+          val r: Kind<F, A>? = when {
+            result.isFailure -> {
+              val e = result.exceptionOrNull()
+              if (e is ShortCircuit) e.recover() else null
```
aballano, line 44: If you want to do a one-liner

```suggestion
              (result.exceptionOrNull() as? ShortCircuit)?.recover()
```

arrow-core-data/src/test/kotlin/arrow/core/EitherTest.kt
```diff
@@ -203,5 +210,83 @@ class EitherTest : UnitSpec() {
...
+    "suspended Either.fx can safely handle immediate exceptions" {
+      Gen.bind(Gen.int(), Gen.throwable(), ::Pair)
+        .random()
+        .take(1001)
+        .forEach { (i, exception) ->
+          shouldThrow<Throwable> {
+            Either.fx<String, Int> {
+              val res = !Either.Right(i)
+              throw exception
+              res
```
aballano, line 251: We could consider testing a case where we perform a side effect here that should never happen to verify that the execution is actually stopped, what do you think?

arrow-core-data/src/test/kotlin/arrow/core/EitherTest.kt
```diff
@@ -203,5 +210,83 @@ class EitherTest : UnitSpec() {
...
+    "suspended Either.fx can bind suspended exceptions" {
+      Gen.bind(Gen.int(), Gen.throwable(), ::Pair)
+        .random()
+        .take(1001)
+        .forEach { (i, exception) ->
+          shouldThrow<Throwable> {
+            Either.fx<String, Int> {
+              val res = !Either.Right(i)
+              sleep(1)
+              throw exception
```
aballano, line 266: I don't see the difference with the test above except for the sleep, which I'm unsure how would change the result. Shouldn't we test the Left binding instead? 🤔

nomisRev: You need to test function that immediately return, and functions that `COROUTINE_SUSPENDED` before they return.

`sleep(1)` guarantees that `COROUTINE_SUSPENDED` is returned [here](https://github.com/arrow-kt/arrow-core/pull/112/files#diff-ffdbb5ceda6c9cb4fc2a0f86a81eaddfR1111).

We test both `Left` and `Right` since we generate `Either`, immediately bind and check that the operation is the same as `identity`.

arrow-core-data/src/main/kotlin/arrow/core/SuspendingMonadContinuation.kt
```diff
@@ -0,0 +1,77 @@
...
+          when {
+            r == null -> {
+              parent.resumeWithException(result.exceptionOrNull()!!)
+              return
+            }
+            _decision.compareAndSet(UNDECIDED, r) -> return
+            else -> Unit // loop again
+          }
+        }
+        else -> { // If not `UNDECIDED` then wwe need to pass result to `parent`
```
pakoito, line 57: typo `we`

arrow-core-data/src/test/kotlin/arrow/core/EitherTest.kt
```diff
@@ -203,5 +211,85 @@ class EitherTest : UnitSpec() {
...
+
+    "suspended Either.fx can bind suspended exceptions" {
+      Gen.bind(Gen.int(), Gen.throwable(), ::Pair)
+        .random()
+        .take(1001)
+        .forEach { (i, exception) ->
+          shouldThrow<Throwable> {
+            Either.fx<String, Int> {
+              val res = !Either.Right(i)
+              sleep(1)
```
pakoito, line 267: 1ms may not be enough to trigger something. It'll make tests slower. Better to have fewer reliable tests.

nomisRev: I used 1 `milliseconds` since that's what I've seen work well :D
`ScheduledThreadPoolExecutor` ensures it gets scheduled to resume in `1000` nanos, no that should make `startCoroutineUninterceptedOrReturn` return `COROUTINE_SUSPENDED`.

I'm not sure if there is a better way to ensure suspension here.

So would you be in favor of reducing iterations to `10` with a `100ms` sleep or `100` with an `20ms` sleep instead?

nomisRev:
@pakoito @aballano I applied all suggested changes, feel free to leave more remarks here. I'm raising another PR updating the current non-suspend version.

PR #124, opened 2020-05-22
Either concrete non-suspend fx

nomisRev:
This is a small PR that implements the same runtime as #112 but in a strict fashion for strict monads that are found in `Arrow-Core`.

This way they can be implemented without the trick currently used in the `FxMonad` implementation. This will unbind us from the JVM, and it will remove all penalties paid by the trick.
This can be kept as a MPP implementation in the future as well for people not opting in for the compiler plugin.

Discussion: `suspend` and `non-suspend` function have a function name clash. This cannot be solved with `@JvmName` since Kotlin cannot infer if it needs the `suspend` or `non-suspend`.


arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -901,6 +900,21 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
       } catch (t: Throwable) {
         fe(t.nonFatalOrThrow()).left()
       }
+
+    fun <E, A> fx2(c: suspend EagerBind<EitherPartialOf<E>>.() -> A): Either<E, A> {
```
pakoito, line 904: `fxEither` for consistency!

pakoito, line 904: `fxEager`? `fxNow`? `fxJust`?

aballano: I'd like to merge as soon as the build is green to continue working with other tickets, so I created an issue to discuss the naming in order not to block us :) Please feel free to continue the discussion there: https://github.com/arrow-kt/arrow-core/issues/133

PR #125, opened 2020-05-22
Fix missed `core/` path in README.md files

hadilq:


pakoito:
Let's wait for a successful build for the docs

PR #129, opened 2020-05-26
Implement `filterIsInstance` and `traverseFilterIsInstance`

hadilq:
Problems:
 - `arrow-meta` cannot handle `java.lang.Class` correctly. It generates `kotlin.Class` which doesn't exist! I tried to add `.replace("java.lang.Class", "")` in `fun String.asKotlin(): String` but it didn't solve the problem! Also in `fun TypeName.Classy.asKotlin(): TypeName.Classy` there is a condition for `Iterable` but didn't work with `Class`!
 - Also I tried with `kotlin.reflect.KClass<T : Any>` but `arrow-meta` cannot handle the `T : Any` part!
we may need to create another PR to solve it then merge it! However, I create this PR to have a reference.


arrow-core-data/src/main/kotlin/arrow/typeclasses/TraverseFilter.kt
```diff
@@ -44,4 +44,12 @@ interface TraverseFilter<F> : Traverse<F>, FunctorFilter<F> {
 
   override fun <A> Kind<F, A>.filter(f: (A) -> Boolean): Kind<F, A> =
     filterA({ Id(f(it)) }, IdApplicative).value()
+
+  /**
+   * Filter out instances of [B] type and traverse the [G] context.
+   */
+  fun <G, A, B> Kind<F, A>.traverseFilterIsInstance(AP: Applicative<G>, klass: Class<B>): Kind<G, Kind<F, B>> = AP.run {
```
pakoito, line 51: One pattern here to prevent some issues we've seen exclusively on Android is to have the `run` only at the minimal scope required.

```
filterA({ a -> AP.just(klass.isInstance(a)) }, AP)
      .map { fa -> AP.run { fa.map { a -> klass.cast(a) } } }
```
```

pakoito:
Thanks @hadilq for your contributions. The changes to arrow-meta will be useful too.

PR #130, opened 2020-05-26
Fix asKotlin extension function to just working with fqName

hadilq:
In https://github.com/arrow-kt/arrow-core/pull/129 PR there is a problem that `arrow-meta` converts `java.lang.Class` to `kotlin.Class`, which does not exist. This is happening because the `asKotlin` extension function tries to convert the packageName alone. In this PR both packageName and simpleName are depending to fqName and `asKotlin` just converts fqName without generally guessing the corresponding kotlin dual of classes.
This solution can have more impact if I would removed all the packageName.asKotlin() and simpleName.asKotlin() in the code, but I would do that if this approach be acceptable one by reviewers.

### DIFF
--- a/arrow-meta/src/main/java/arrow/meta/encoder/jvm/StringTypeExtensions.kt
+++ b/arrow-meta/src/main/java/arrow/meta/encoder/jvm/StringTypeExtensions.kt
@@ -28,13 +28,10 @@ fun String.asKotlin(): String =
     .replace("java.util.SortedMap", "kotlin.collections.SortedMap")
     .replace("java.util.Collection", "kotlin.collections.Collection")
     .replace("java.lang.Number", "kotlin.Number")
-    .replace("java.lang.Throwable", "kotlin.Throwable").let {
-      if (it == "java.lang") it.replace("java.lang", "kotlin")
-      else it
-    }.let {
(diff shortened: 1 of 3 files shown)


arrow-meta/src/main/java/arrow/meta/encoder/jvm/StringTypeExtensions.kt
```diff
@@ -28,13 +28,10 @@ fun String.asKotlin(): String =
...
-    }.let {
-      if (it == "java.util") it.replace("java.util", "kotlin.collections")
-      else it
-    }
+    .replace("java.lang.Comparable", "kotlin.Comparable")
+    .replace("java.lang.Boolean", "kotlin.Boolean")
+    .replace("java.lang.Long", "kotlin.Long")
+    .replace("java.lang.Throwable", "kotlin.Throwable")
     .replace("kotlin.Integer", "kotlin.Int")
     .replace("Integer", "Int")
```
aballano, line 36: I believe, because you removed this weird let from above, that this should be 

```suggestion
    .replace("java.lang.Integer", "kotlin.Int")
```

raulraja:
thanks @hadilq !

Issue #116, opened 2020-05-21
Support `suspend` in `fx`


nomisRev:
Currently we don't support `suspend` for `fx` blocks.

This prevents mixing `suspend` with `fx`, which would be the most Kotlin idiomatic.

```
import arrow.core.Either
import arrow.core.extensions.fx

data class User(val name: String)
object DomainError

suspend fun fetchDataFromNetwork(id: Int): Either<DomainError, User> = TODO()

suspend fun getUser(id: Int) : Either<DomainError, User> =
  Either.fx<DomainError, User> {
    // Restricted suspending functions can only invoke member or extension suspending functions on their restricted coroutine scope
    !fetchDataFromNetwork(id) 
  }
```

In #112, this change was implemented for `Either`, and it can be implemented for all other data types in the same way.

This needs to happen for all data types that are not marked for deprecation in #114.

 - [x] `Either`
 - [x] `Validated`
 - [ ] `NonEmptyList`
 - [x] `Eval`
 - [x] `Const`
 - [ ] `Ior`

aballano:
All of NonEmptyList, ListK, SequenceK and Ior would need a rework on the current system to use flatmap (or something similar) as otherwise we cannot achieve the same operation with the current architecture.

aballano:
Leftover datatypes will be tackled in #171

PR #131, opened 2020-05-27
Suspend Validated.fx implementation

aballano:
## Done in this PR

* Add suspend fx for Validated
* Add Validated.fromNullable
* Fix an issue on Validated docs for error accumulation
* Rework Validated docs to include an fx example

## Details

As per #116 this PR **adds fx to validated**, so one simple example can be found in the **docs of validated** itself, where we have the following code:

```kotlin
data class Config(val map: Map<String, String>) {
 fun <A> parse(read: Read<A>, key: String): Validated<ConfigError, A> {
  val v = Option.fromNullable(map[key])
  return when (v) {
   is Some ->
    when (val s = read.read(v.t)) {
     is Some -> s.t.valid()
     is None -> ConfigError.ParseConfig(key).invalid()
    }
   is None -> Validated.Invalid(ConfigError.MissingConfig(key))
  }
 }
}
```

So, instead of dealing with option, **I added a `Validated.fromNullable`** to match `fromOption` already present and simply deal with validates in an fx block:

```kotlin
data class Config(val map: Map<String, String>) {
  suspend fun <A> parse(read: Read<A>, key: String) = Validated.fx<ConfigError, A> {
    val value = Validated.fromNullable(map[key]) {
      ConfigError.MissingConfig(key)
    }.bind()

    val readVal = Validated.fromOption(read.read(value)) {
      ConfigError.ParseConfig(key)
    }.bind()

    readVal
  }
}
```

The only difference is that we force `parse` to be **suspended** because of this version of fx being suspended (shall we wait for #124?).


arrow-core-data/src/main/kotlin/arrow/core/Validated.kt
```diff
@@ -217,29 +250,28 @@ typealias Invalid<E> = Validated.Invalid<E>
...
+ *  when {
  *   v1 is Validated.Valid && v2 is Validated.Valid -> Validated.Valid(f(v1.a, v2.a))
- *   v1 is Validated.Valid && v2 is Validated.Invalid -> v2
- *   v1 is Validated.Invalid && v2 is Validated.Valid -> v1
- *   v1 is Validated.Invalid && v2 is Validated.Invalid -> TODO()
- *   else -> TODO()
+ *   v1 is Validated.Valid && v2 is Validated.Invalid -> v2.toValidatedNel()
+ *   v1 is Validated.Invalid && v2 is Validated.Valid -> v1.toValidatedNel()
+ *   v1 is Validated.Invalid && v2 is Validated.Invalid -> Validated.Invalid(NonEmptyList(v1.e, listOf(v2.e)))
+ *   else -> throw IllegalStateException("Not possible value")
```
pakoito, line 271: Instead of `Throw`, we could return an Invalid with it.

aballano: But that's the impossible case, does it really matter? Actually I was looking for if there was an operator that does this already and doesn't leave the burden of writing this to the dev but couldn't see any

pakoito: Isn't it exhaustive then?

aballano: It cannot be because we're checking 2 elements

pakoito: we need pattern matching *weep*

aballano: Cannot we add a mapN-like function to do this already? This way we could hide the throwing under our own impl cc @nomisRev

nomisRev: Yes, you can use `Applicative` for this. 

```
Validated.applicative(NonEmptyList.semigroup<String>()).tupledN(
  1.valid(),
  "2".invalidNel(),
  "3".invalidNel()
) // Validated.Invalid(NonEmptyList("2", "3"))
```

aballano: Right, I just realized that we're in core-data and we don't have access to the extensions :/

What I did is the following:

* Added a convenient `nelApplicative` function in Validated extensions
* Suggest to use this approach instead of the custom function in docs, but without changing the examples code due to the import problem

aballano:
@pakoito Would you take a second look at the docs when you have time, please? 🙏😁

arrow-core-data/src/main/kotlin/arrow/core/Validated.kt
```diff
@@ -133,6 +136,32 @@ typealias Invalid<E> = Validated.Invalid<E>
  * //sampleEnd
  * ```
  *
+ * But, as you can see, the parser runs sequentially: first tries to get the map value and then tries to read it.
```
pakoito, line 139: `But` -> `and`. Just so you get the habit :D

arrow-core-data/src/main/kotlin/arrow/core/Validated.kt
```diff
@@ -133,6 +136,32 @@ typealias Invalid<E> = Validated.Invalid<E>
...
+ * This is easily translated to an Fx block:
```
pakoito, line 140: If you wanna get picky on wording, `easy` -> `straightforward` or `directly`.

arrow-core-data/src/main/kotlin/arrow/core/Validated.kt
```diff
@@ -179,9 +208,39 @@ typealias Invalid<E> = Validated.Invalid<E>
...
+ *
+ * ```kotlin:ank:silent
+ * import arrow.core.NonEmptyList
+ * import arrow.core.Validated
+ * import arrow.core.ValidatedPartialOf
+ * import arrow.core.extensions.nonemptylist.semigroup.semigroup
+ * import arrow.core.extensions.validated.applicative.applicative
+ * import arrow.core.fix
+ * import arrow.typeclasses.Applicative
+ * // added manually due to deps
```
pakoito, line 226: Additional white line?

arrow-core-data/src/main/kotlin/arrow/core/Validated.kt
```diff
@@ -618,19 +674,41 @@ sealed class Validated<out E, out A> : ValidatedOf<E, A> {
...
     /**
-     * Converts an `Either<A, B>` to an `Validated<A, B>`.
+     * Converts an `Either<E, A>` to a `Validated<E, A>`.
      */
     fun <E, A> fromEither(e: Either<E, A>): Validated<E, A> = e.fold({ Invalid(it) }, { Valid(it) })
 
     /**
-     * Converts an `Option<B>` to an `Validated<A, B>`, where the provided `ifNone` values is returned on
-     * the invalid of the `Validated` when the specified `Option` is `None`.
+     * Converts an `Option<A>` to a `Validated<E, A>`, where the provided `ifNone` output value is returned as invalid
```
pakoito, line 682: `[Invalid]`?

arrow-core-data/src/main/kotlin/arrow/core/Validated.kt
```diff
@@ -363,18 +420,17 @@ typealias Invalid<E> = Validated.Invalid<E>
...
- *   val v = Option.fromNullable(map[key])
- *   return when (v) {
- *    is Some ->
- *     when (val s = read.read(v.t)) {
- *      is Some -> s.t.valid()
- *      is None -> ConfigError.ParseConfig(key).invalid()
- *     }
- *    is None -> Validated.Invalid(ConfigError.MissingConfig(key))
+ *   suspend fun <A> parse(read: Read<A>, key: String) = Validated.fx<ConfigError, A> {
+ *     val value = Validated.fromNullable(map[key]) {
```
pakoito, line 424: Neat

arrow-core-data/src/main/kotlin/arrow/core/Validated.kt
```diff
@@ -363,18 +420,17 @@ typealias Invalid<E> = Validated.Invalid<E>
...
+ *       ConfigError.MissingConfig(key)
+ *     }.bind()
+ *     val readVal = Validated.fromOption(read.read(value)) {
```
pakoito, line 427: Great! Good use of both APIs

arrow-core-data/src/main/kotlin/arrow/core/Validated.kt
```diff
@@ -388,11 +444,11 @@ typealias Invalid<E> = Validated.Invalid<E>
  *
  * val config = Config(mapOf("house_number" to "-42"))
  *
- * val houseNumber = config.parse(Read.intRead, "house_number").withEither { either ->
- *  either.flatMap { positive("house_number", it) }
- * }
+ * suspend fun main() {
+ *   val houseNumber = config.parse(Read.intRead, "house_number").withEither { either ->
+ *    either.flatMap { positive("house_number", it) }
```
pakoito, line 449: one additional space

arrow-core-data/src/main/kotlin/arrow/core/Validated.kt
pakoito: It's just meh in general. It could be possible to rewrite it as two nested folds and some manual formatting.

arrow-core-data/src/main/kotlin/arrow/core/Validated.kt
nomisRev: I'm not sure I understand why this is necessary? Can we not use `Validated.applicative(NonEmptyList.semigroup<String>()).tupledN` in the ank snippet?

arrow-core/src/main/kotlin/arrow/core/extensions/validated.kt
```diff
@@ -185,3 +188,6 @@ interface ValidatedHash<L, R> : Hash<Validated<L, R>>, ValidatedEq<L, R> {
     HR().run { it.hash() }
   })
 }
+
+fun <E> Validated.Companion.nelApplicative(): ValidatedApplicative<NonEmptyList<E>> =
```
nomisRev, line 192: Shouldn't it be `applicativeNel`, inline with `ValidatedNel`? That way people will also get this auto suggested if they start typing `Validated.app..`.

arrow-core-data/src/main/kotlin/arrow/core/Validated.kt
nomisRev: I'm not a fan of showing `parallelValidate` at all.. It promotes/shows code that no-one should ever write using Arrow.

We have a bunch of `helpers` in the `docs` module, any way we can provide instances there to make the desired snippets compile?

arrow-core-data/src/main/kotlin/arrow/core/Validated.kt
aballano: Yes, we're actually using that already, I just wanted a quick way of obtaining it for any E

arrow-core/src/main/kotlin/arrow/core/extensions/validated.kt
aballano: derp, good point! 😄

arrow-core-data/src/main/kotlin/arrow/core/Validated.kt
aballano: The only reason I like it is for the explanation vs a Validated Nel, which is also explained in there. I changed it to try to make it clear that this is just for the shake of the example

aballano:
@nomisRev If you have any more comments later please let me know so I can open another PR :)

PR #142, opened 2020-06-05
Fix a few broken links in the docs

cb372:
I couldn't work out exactly why they were broken. Tried a few different
things but couldn't get it to generate the link correctly. Ended up
removing the curly braces and using a plain old Markdown link.

Also removed the "type class hierarchy" section from these pages. They
are not type classes, so the macro was generating an empty diagram.

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
@@ -11,8 +11,6 @@ import kotlin.coroutines.intrinsics.suspendCoroutineUninterceptedOrReturn
 
 /**
  *
- * ank_macro_hierarchy(arrow.core.Either)
- *
  *
  * In day-to-day programming, it is fairly common to find ourselves writing functions that can fail.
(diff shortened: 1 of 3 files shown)


pakoito:
I believe we don't use relative paths at all. They're broken because the docs were reparented to other paths instead.

rachelcarmena:
> I believe we don't use relative paths at all. They're broken because the docs were reparented to other paths instead.

They are used however I think they were wrong. I'll try it asap.

rachelcarmena:
> @rachelcarmena Thanks for the review!
> 
> I considered doing what you are proposing in your suggested changes, i.e. linking to the KDoc pages. But I thought it might be better to link to the hand-written typeclass explanation pages? (I think that's what the broken links are _trying_ to link to.) I'm happy to go with whatever you prefer.

Oh, sorry, I see. In that case I think it's not possible to use `{{ }}`  because it's not considered as a link by Dokka and transforms it in a `(#)` link.

Thanks @cb372 !!! :raised_hands:

PR #143, opened 2020-06-05
Replace compile with implementation in setup guide

LordRaydenMK:
From the gradle docs:
The compile configuration still exists but should not be used as it will not offer the guarantees that the api and implementation configurations provide.

https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation

I was replacing `compile` with `implementation` when creating a new project, after copy-pasting from the docs, and today I decided it's the last time I do it :smile:


pakoito:
LGTM. Another pair of eyes that does Android regularly? @JorgeCastilloPrz ?

PR #148, opened 2020-06-09
Fix Mapk extensions

### LINKED ISSUE #140: ["Request"] Fix Map instances generation
Having this example:

```kotlin
val m = mapOf("a" to listOf(1, 2).k(), "b" to listOf(3).k()).k()
val n = mapOf("a" to listOf(4).k(), "c" to listOf(5, 6).k()).k()
val p = MapK.semigroup<String, ListK<Int>>(ListK.semigroup()).run { m + n }
```
// Where is p -> `Map([(a, [4, 1, 2]), (b, [3]), (c, [5, 6])])` as expected(?) (doing m+n i would expect a to be 1,2,4 tho)

There are several problems that I can extract:

* Need to apply `.k()` everywhere. Because `MapK` is a `Map` and `ListK` is a `List`, we should be able to generate the semigroup instance for them too, and actually we can for List, but Map doesn't get extensions for some reason.
* Removing any `.k()` doesn't make the compilation to fail in any way but rather produces a different output because it defaults to Iterable.plus implementation from the stdlib. The output is then -> `{a=[4], b=[3], c=[5, 6]}`

aballano:
* Added `listKOf` for ListK as equivalent to List's `listOf`
  -> With this, there's no need to create normal Lists and Maps to then convert them to the wrapped version via `.k()`, but instead just create the wrapped versions directly as follows:

```
val m = mapOf("a" toT  listKOf(1, 2), "b" toT listKOf(3)) // Note that the `toT` will use an Arrow's Tuple to create a MapK
val n = mapOf("a" toT listKOf(4), "c" toT listKOf(5, 6))
val p = m.plus(ListK.semigroup(), n)
```

* Fix MapK extensions projected over stdlib Map
  -> So now there's no need to convert all maps to MapK because we can use the Map extensions as follows (note that we cannot still use `List.semigroup` tho):

```
val m = mapOf("a" to 1, "b" to 3)
val n = mapOf("a" to 5, "c" to 5)
val p = m.plus(Int.semigroup(), n)
```

Fixes #140

### DIFF
--- a/arrow-meta/src/main/java/arrow/meta/encoder/jvm/JvmMetaApi.kt
+++ b/arrow-meta/src/main/java/arrow/meta/encoder/jvm/JvmMetaApi.kt
@@ -600,7 +600,7 @@ interface JvmMetaApi : MetaApi, TypeElementEncoder, ProcessorUtils, TypeDecoder 
    * ex: SetK<A> to Set<A>
    */
   val Type.kindWrapper: Pair<TypeName, TypeName.ParameterizedType>?
-    get() = if (primaryConstructor?.parameters?.size == 1 && typeVariables.size == 1) {
+    get() = if (primaryConstructor?.parameters?.size == 1) {
       val wrappedType = primaryConstructor.parameters[0].type.asKotlin()
       when (wrappedType) {
         is TypeName.ParameterizedType -> {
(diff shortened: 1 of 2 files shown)


arrow-meta/src/main/java/arrow/meta/encoder/jvm/JvmMetaApi.kt
```diff
@@ -600,7 +600,7 @@ interface JvmMetaApi : MetaApi, TypeElementEncoder, ProcessorUtils, TypeDecoder
    * ex: SetK<A> to Set<A>
    */
   val Type.kindWrapper: Pair<TypeName, TypeName.ParameterizedType>?
-    get() = if (primaryConstructor?.parameters?.size == 1 && typeVariables.size == 1) {
+    get() = if (primaryConstructor?.parameters?.size == 1) {
```
pakoito, line 603: ez pz

aballano, line 603: @raulraja Hope this is ok tho 😅

PR #153, opened 2020-06-12
Enable incremental annotation processing

hadilq:
This PR is created after https://github.com/arrow-kt/arrow/issues/2032 issue to enable incremental annotation processing. The following is the list of assumptions and changes.

 - Update `auto-service` to `1.0-rc7` where based on https://github.com/google/auto/issues/615 must support incremental annotation processing. We may need to create a PR for https://github.com/arrow-kt/arrow to update the `GOOGLE_AUTO_SERVICE_VERSION` value,
 - Use `File` instead of `Filer` when the `packageName` contains Java keywords like `const` or `try`
 - Ignore double attempt to generate a file where throws `FilerException: Attempt to reopen a file for path ...`
 - We may need to create PRs for other Arrow projects to add `kapt.incremental.apt=true` line to `gradle.properties` files.
 - Additionally, there are some fix of indention, where just happened when I reformat working files in this PR


arrow-meta/arrow-meta-test-models/build.gradle
```diff
@@ -10,8 +10,8 @@ apply from: "$DOC_CONF"
 
 dependencies {
     compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$KOTLIN_VERSION"
-    compileOnly "com.google.auto.service:auto-service:$GOOGLE_AUTO_SERVICE_VERSION"
-    kapt "com.google.auto.service:auto-service:$GOOGLE_AUTO_SERVICE_VERSION"
+    compileOnly "com.google.auto.service:auto-service:1.0-rc7"
+    kapt "com.google.auto.service:auto-service:1.0-rc7"
```
rachelcarmena, line 14: Thanks for this awesome PR @hadilq :raised_hands: I was thinking about this kind of changes and I think the mistake was to locate all the versions in a centralized place without filtering them.

I mean, in this case, `GOOGLE_AUTO_SERVICE_VERSION` (and `GOOGLE_COMPILE_TESTING_VERSION` as well) is just being used in this repository so I think it doesn't make sense to be placed in `arrow` repository.

So please, feel free to locate those versions in this repository and remove them from `arrow` repository. 

I think `arrow` repository should contain versions that are relevant and can impact to all the libraries. What do you think about it?

Thanks again @hadilq !!

hadilq: Thank you @rachelcarmena for your this great suggestion. I'm totally agree and you can find the changes in b8d1703 and https://github.com/arrow-kt/arrow/pull/2158

rachelcarmena: Awesome, thanks @hadilq for your contribution!!

hadilq:
Hi everyone! It took more than two weeks to review it. Please feel free to give me some feedback. Thanks :)

PR #154, opened 2020-06-16
Suspend Eval.fx implementation

aballano:
Also created FxLaws


arrow-core-test/src/main/kotlin/arrow/core/test/laws/FxLaws.kt
```diff
@@ -0,0 +1,166 @@
...
+    Law("non-suspended fx can bind immediate exceptions") { nonSuspendedCanBindImmediateException(pureGen, fxBlock) },
+    Law("suspended fx can bind immediate values") { suspendedCanBindImmediateValues(G, sfxBlock) },
+    Law("suspended fx can bind suspended values") { suspendedCanBindSuspendedValues(G, sfxBlock) },
+    Law("suspended fx can bind immediate exceptions") { suspendedCanBindImmediateExceptions(pureGen, sfxBlock) },
+    Law("suspended fx can bind suspended exceptions") { suspendedCanBindSuspendedExceptions(pureGen, sfxBlock) }
+  )
+
+  private suspend fun <F, A> nonSuspendedIsLazy(G: Gen<Kind<F, A>>, fxBlock: EagerFxBlock<F, A>) {
+    forAll(G) { f: Kind<F, A> ->
+      val effect = SideEffect()
```
aballano, line 42: @nomisRev I'm unsure if this is expected, but all new fx blocks are not lazy by default it seems. Or I'm doing smth wrong :/

nomisRev: All the data types except for Eval are eager. Which we can make eager by rewrapping in `Eval`, which is what we do now too I think? 🤔

aballano: Right, it was only eval which we made lazy + IO 👌 I'll remove this tests then and leave Eval as an special case, although it might be confusing for people who don't know which types can be lazy or not, maybe it would make sense to typealias the syntax for them to LazySyntax (similarly as EagerSyntax now)?. For example I'm thinking on SequenceK here 🤔

nomisRev: `fx` is a DSL to avoid nesting with `flatMap`, so the users should still be aware of what they're doing with the actual data types.

`SequenceK` will require `flatMap` rewrite afaik, or a multi-shot continuation solution.

arrow-core-test/src/main/kotlin/arrow/core/test/laws/FxLaws.kt
```diff
@@ -0,0 +1,166 @@
...
+      fxBlock {
+        effect.increment()
+        f.bind()
+      }
+
+      effect.counter == 0
```
nomisRev, line 48: This should be `1` for an eager block, no?

arrow-core-test/src/main/kotlin/arrow/core/test/laws/FxLaws.kt
```diff
@@ -0,0 +1,166 @@
...
+      }
+  }
+
+  private suspend fun <F, A> suspendedCanBindSuspendedValues(G: Gen<Kind<F, A>>, fxBlock: SuspendFxBlock<F, A>) {
+    G.random()
+      .take(10)
+      .forEach { f ->
+        fxBlock {
+          val res = !(suspend {
+            sleep(100)
```
nomisRev, line 106: I wrote a set of generic functions `suspended()` that allow you to lift an immediate value, or `Throwable` to a suspended value or exception without having to sleep.

https://github.com/arrow-kt/arrow-fx/blob/master/arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/predef-test.kt#L123

arrow-core-test/src/main/kotlin/arrow/core/test/laws/FxLaws.kt
```diff
@@ -0,0 +1,166 @@
...
+  private suspend fun <F, A> suspendedCanBindSuspendedExceptions(G: Gen<Kind<F, A>>, fxBlock: SuspendFxBlock<F, A>) {
+    Gen.bind(G, Gen.throwable(), ::Pair)
+      .random()
+      .take(10)
+      .forEach { (f, exception) ->
+        shouldThrow<Throwable> {
+          fxBlock {
+            val res = !f
+            sleep(100)
+            throw exception
```
nomisRev, line 140: Lift into `suspend` without `sleep`.

arrow-core-test/src/main/kotlin/arrow/core/test/laws/FxLaws.kt
aballano: I removed the lazy tests, basically because the suspended fx block is lazy by default, and having the non-suspend one to be lazy will cause issues with the other tests, so better leave like this I think

Issue #133, opened 2020-05-29
Figure out a name for non-suspend fx block


aballano:
Related to #117

> `fxEager`? `fxNow`? `fxJust`?

_Originally posted by @pakoito in https://github.com/arrow-kt/arrow-core/pull/124_

pakoito:
Oh thank god please yes

aballano:
I'm in favour of `fxEager` or `eagerFx`

pakoito:
fxEager for discoverability

aballano:
Seems that the [problem is solved(?)](https://github.com/arrow-kt/arrow-core/pull/160/files#diff-048837e2615e457be422388de720353fR46) 👀

Issue #117, opened 2020-05-21
Refactor existing non-suspend fx blocks


nomisRev:
The current `fx` implementation is coupled to the JVM, and this prevents Arrow-Core from becoming MPP.

The new implementation in #112 can be adjusted to work with `@RestrictSupension`, and not be coupled to the JVM.

The new implementation should also be exposed from `arrow-core-data` on the respective `companion object`.

This needs to happen for all data types that are not marked for deprecation in #114.

 - [x] `Either`
 - [x] `Validated`
 - [ ] `NonEmptyList`
 - [x] `Eval`
 - [x] `Const`
 - [ ] `Ior`

A reference PR will be linked asap.

aballano:
Leftover datatypes will be tackled in #171

PR #160, opened 2020-06-18
Suspend Const.fx implementation

aballano:
Also changes all fx blocks to match their datatype as discussed in #133

Related to #116 #117


aballano:
> We should also deprecate the old fx blocks, but we can do that separately later.

Yes, it makes sense. I think I deleted them all, so next think I'll raise a PR bringing them back and deprecating them 👌 

Do we also want to deprecate Monad.fx in the process?

Issue #115, opened 2020-05-21
Properly support `suspend` with `inline`


nomisRev:
Currently, it's hard to combine `suspend` with the data types found in `Arrow Core`.

This hinders from writing idiomatic functional code in Kotlin.

```kotlin
import arrow.core.Either
import arrow.core.flatMap

data class User(val name: String)
object DomainError

suspend fun fetchDataFromNetwork(id: Int): Either<DomainError, User> = TODO()

suspend fun program(): Either<DomainError, User> =
  Either.Right(1).flatMap { id ->
    fetchDataFromNetwork(id) // Suspension functions can be called only within coroutine body
  }
```

However, if we refactor `flatMap` to be `inline` it'll be allowed.

`inline fun <A, B, C> EitherOf<A, B>.flatMap(f: (B) -> Either<A, C>): Either<A, C>`

This is in conflict with #69 which proposes a duplicated `suspend` API.

This needs to happen for all data types that are not marked for deprecation in #114.

 - [x] `Either` #163 by @worldsnas
 - [x] `Validated`
 - [x] `NonEmptyList`
 - [ ] `Eval`
 - [x] `Const`
 - [x] `Ior`

pakoito:
I believe this could be resolved with `suspend` versions of the functions instead. If we want to support it, let's do it explicitly.

`inline` allows suspension because it permits non-local returns. This is an anti-pattern for purity purposes, and can lead to very subtle bugs. It's an unintended coincidence that it's a scape hatch inside suspend functions.

Secondly, it causes bytecode bloat. Instead of having a single function on the data part of your binary, you have hundreds or thousands, each the same. This is bad for startup time and download and install time of apps.

Lastly, the same good thing about inline copypasting the code leads to runtime exceptions by bad code. Inlining, same as text macros, can and will produce bad bytecode on colliding definitions.

I don't believe we should normalize or encourage this just for a bit of adoption. Having a second `suspend` set of functions solves the same problem for cases where people want to call suspend functions inside those blocks.

Existing art:
https://docs.hhvm.com/hsl/reference/function/HH.Lib.Vec.map/
https://docs.hhvm.com/hsl/reference/function/HH.Lib.Vec.map_async/

https://docs.rs/futures/0.3.5/futures/stream/trait.StreamExt.html#method.for_each
https://docs.rs/futures/0.3.5/futures/stream/trait.StreamExt.html#method.for_each_concurrent

nomisRev:
[...]

PR #163, opened 2020-06-29
Inlining Either operators

worldsnas:
Inlining operators allows me to use local suspend functions inside them.

One use case for me was to use a simple database call inside `flatMap` but since it was not inline I could not.

Please let me if I missed any guidelines or contribution policy


aballano:
Thank you for the contribution @worldsnas! 🚀 This partially fixes https://github.com/arrow-kt/arrow-core/issues/115 for Either 🎉

PR #167, opened 2020-06-30
Deprecate `fx` blocks from typeclasses

### LINKED ISSUE #118: Deprecate `fx` blocks from typeclasses
We're going to provide top-level `computation` blocks in a `suspend` and `non-suspend` style in #116 & #117.

When Arrow Meta is introduced then we will remove `fx` from the type classes and a compiler plugin will be provided instead to structurally enable `non-suspend` `fx` on `flatMap`.

If a `suspend` fx version can be provided by the compiler plugin, it will also provide it.

aballano:
Deprecate Monad.fx 
Deprecates fx blocks in Either and Eval

Fixes #118


aballano:
> The data types we're deprecated don't have to be deprecated further. And `Ior`, `List`, etc don't have their concrete replacement yet, so this doesn't fix #118 yet.

Right, I created https://github.com/arrow-kt/arrow-core/issues/171 to tackle those specifically, so we can keep moving forward here :)

PR #175, opened 2020-07-06
Adds Validated.catch

JorgeCastilloPrz:
This mimics `Either.catch` constructor for `suspend` ops but for `Validated` and `ValidatedNel`. I wrote one of these manually [in this Ank PR](https://github.com/arrow-kt/arrow-ank/pull/43), so decided to add it to `arrow-core`.


JorgeCastilloPrz:
I think this is failing because of the unrelated `IO` revert status.

nomisRev:
@JorgeCastilloPrz Yes, still broken due to wrong imports in some tutorial-like docs again..

PR #180, opened 2020-07-06
Inline datatypes

aballano:
Related to #115

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/core/Const.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Const.kt
@@ -15,33 +15,43 @@ fun <A, T> ConstOf<A, T>.value(): A = this.fix().value()
 data class Const<A, out T>(private val value: A) : ConstOf<A, T> {
 
   @Suppress("UNCHECKED_CAST")
-  fun <U> retag(): Const<A, U> = this as Const<A, U>
+  fun <U> retag(): Const<A, U> =
+    this as Const<A, U>
 
   @Suppress("UNUSED_PARAMETER")
-  fun <G, U> traverse(GA: Applicative<G>, f: (T) -> Kind<G, U>): Kind<G, Const<A, U>> = GA.just(retag())
+  fun <G, U> traverse(GA: Applicative<G>, f: (T) -> Kind<G, U>): Kind<G, Const<A, U>> =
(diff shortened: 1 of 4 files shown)


arrow-core-data/src/main/kotlin/arrow/core/Const.kt
```diff
@@ -15,33 +15,43 @@ fun <A, T> ConstOf<A, T>.value(): A = this.fix().value()
...
+    GA.just(retag())
 
   companion object {
-    fun <A, T> just(a: A): Const<A, T> = Const(a)
+    fun <A, T> just(a: A): Const<A, T> =
+      Const(a)
   }
 
-  fun value(): A = value
+  fun value(): A =
```
aballano, line 34: I'm wondering out loud, wouldn't be simpler to just expose `value` in the first place instead?

nomisRev: This was originally done to expose a `value()` method on both the concrete and kinded `xxxOf` variants.

Perf-wise this doesn't cause any impact one way or the other.

aballano:
> Looks good. Is it the line of thought that everything that uses an `f` needs to be inlined?

that's more or less what I tried, although it cannot be applied to all cases in which you pass the lambda around (like in Eval mainly), so there are some function inlining leftover until we figure it out another way.

arrow-core-data/src/main/kotlin/arrow/core/Ior.kt
```diff
@@ -139,7 +139,8 @@ sealed class Ior<out A, out B> : IorOf<A, B> {
     is Both -> fab(leftValue, rightValue)
   }
 
-  fun <C> foldLeft(c: C, f: (C, B) -> C): C = fold({ c }, { f(c, it) }, { _, b -> f(c, b) })
+  inline fun <C> foldLeft(c: C, f: (C, B) -> C): C =
+    fold({ c }, { f(c, it) }, { _, b -> f(c, b) })
 
   fun <C> foldRight(lc: Eval<C>, f: (B, Eval<C>) -> Eval<C>): Eval<C> =
```
aballano, line 145: @JorgeCastilloPrz this is an example, as you see, because f is passed inside Eval.defer { } and it is stored in a data class, so it cannot be inlined.

PR #192, opened 2020-07-23
Properly support suspend with inline

### LINKED ISSUE #115: Properly support `suspend` with `inline`

aballano:
* Make the remaining functions inline with crossinline lambdas

Also:
* 👀 Remove unused lambda in Const's traverse and traverseFilter
* 🔧 Deprecate Validated.leftMap in favour of mapLeft
* 🔧 Deprecate Either.catch(fe, f) in favour catch(f).mapLeft(fe)
* ✅ Add missing tests to Either.catch

Fixes #115

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/core/Const.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Const.kt
@@ -15,10 +15,12 @@ data class Const<A, out T>(private val value: A) : ConstOf<A, T> {
   fun <U> retag(): Const<A, U> =
     this as Const<A, U>
 
-  fun <G, U> traverse(GA: Applicative<G>): Kind<G, Const<A, U>> =
+  @Suppress("UNUSED_PARAMETER")
+  fun <G, U> traverse(GA: Applicative<G>, f: (T) -> Kind<G, U>): Kind<G, Const<A, U>> =
     GA.just(retag())
 
-  fun <G, U> traverseFilter(GA: Applicative<G>): Kind<G, Const<A, U>> =
+  @Suppress("UNUSED_PARAMETER")
+  fun <G, U> traverseFilter(GA: Applicative<G>, f: (T) -> Kind<G, Option<U>>): Kind<G, Const<A, U>> =
     GA.just(retag())
 
   companion object {
@@ -43,7 +45,7 @@ fun <A, T, U> ConstOf<A, T>.ap(SG: Semigroup<A>, ff: ConstOf<A, (T) -> U>): Cons
   fix().retag<U>().combine(SG, ff.fix().retag())
 
 fun <T, A, G> ConstOf<A, Kind<G, T>>.sequence(GA: Applicative<G>): Kind<G, Const<A, T>> =
-  fix().traverse(GA)
+  fix().traverse(GA, ::identity)
 
 inline fun <A> A.const(): Const<A, Nothing> =
   Const(this)
--- a/arrow-core-data/src/main/kotlin/arrow/core/Const.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Const.kt
@@ -15,12 +15,10 @@ data class Const<A, out T>(private val value: A) : ConstOf<A, T> {
   fun <U> retag(): Const<A, U> =
     this as Const<A, U>
 
-  @Suppress("UNUSED_PARAMETER")
-  fun <G, U> traverse(GA: Applicative<G>, f: (T) -> Kind<G, U>): Kind<G, Const<A, U>> =
+  fun <G, U> traverse(GA: Applicative<G>): Kind<G, Const<A, U>> =
     GA.just(retag())
 
-  @Suppress("UNUSED_PARAMETER")
-  fun <G, U> traverseFilter(GA: Applicative<G>, f: (T) -> Kind<G, Option<U>>): Kind<G, Const<A, U>> =
+  fun <G, U> traverseFilter(GA: Applicative<G>): Kind<G, Const<A, U>> =
     GA.just(retag())
 
   companion object {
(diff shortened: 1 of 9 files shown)


arrow-core-data/src/main/kotlin/arrow/core/Eval.kt
```diff
@@ -244,15 +243,18 @@ sealed class Eval<out A> : EvalOf<A> {
...
+  fun <B> ap(ff: EvalOf<(A) -> B>): Eval<B> =
+    ff.fix().flatMap { f -> map(f) }.fix()
 
   @Suppress("PARAMETER_NAME_CHANGED_ON_OVERRIDE", "UNCHECKED_CAST")
-  fun <B> flatMap(f: (A) -> EvalOf<B>): Eval<B> =
+  inline fun <B> flatMap(crossinline f: (A) -> EvalOf<B>): Eval<B> =
     when (this) {
       is FlatMap<A> -> object : FlatMap<B>() {
         override fun <S> start(): Eval<S> = (this@Eval).start()
+        @IgnoreJRERequirement
```
JorgeCastilloPrz, line 257: Can you clarify the need for this? 🙏

arrow-core-data/src/main/kotlin/arrow/core/Validated.kt
```diff
@@ -770,11 +770,15 @@ sealed class Validated<out E, out A> : ValidatedOf<E, A> {
   inline fun <B> map(f: (A) -> B): Validated<E, B> =
     bimap(::identity, f)
 
+  @Deprecated("Use mapLeft for consistency", ReplaceWith("mapLeft(f)"))
+  inline fun <EE> leftMap(f: (E) -> EE): Validated<EE, A> =
```
JorgeCastilloPrz, line 774: 👍

arrow-core-data/src/test/kotlin/arrow/core/EitherTest.kt
```diff
@@ -216,5 +216,16 @@ class EitherTest : UnitSpec() {
           Right(a).handleErrorWith { Right(b) } == Right(a)
       }
     }
+
+    "catch should return Valid(result) when f does not throw" {
```
JorgeCastilloPrz, line 220: Thanks, we were missing these. I should have added them while working on `Validated.catch {}`, I think I noticed `Either` was missing these but totally forgot to add them in the end 🤦

arrow-core/src/main/kotlin/arrow/core/extensions/const.kt
```diff
@@ -107,7 +107,7 @@ interface ConstTraverse<X> : Traverse<ConstPartialOf<X>>, ConstFoldable<X> {
   override fun <T, U> ConstOf<X, T>.map(f: (T) -> U): Const<X, U> = fix().retag()
 
   override fun <G, A, B> ConstOf<X, A>.traverse(AP: Applicative<G>, f: (A) -> Kind<G, B>): Kind<G, ConstOf<X, B>> =
-    fix().traverse(AP, f)
+    fix().traverse(AP)
```
JorgeCastilloPrz, line 110: Not removing the param from the declaration also here? should we deprecate it at least?

arrow-core/src/main/kotlin/arrow/core/extensions/const.kt
```diff
@@ -116,7 +116,7 @@ interface ConstTraverseFilter<X> : TraverseFilter<ConstPartialOf<X>>, ConstTrave
   override fun <T, U> Kind<ConstPartialOf<X>, T>.map(f: (T) -> U): Const<X, U> = fix().retag()
 
   override fun <G, A, B> Kind<ConstPartialOf<X>, A>.traverseFilter(AP: Applicative<G>, f: (A) -> Kind<G, Option<B>>): Kind<G, ConstOf<X, B>> =
-    fix().traverseFilter(AP, f)
+    fix().traverseFilter(AP)
```
JorgeCastilloPrz, line 119: Same here, feels confusing to have an unused parameter there.

arrow-core-data/src/main/kotlin/arrow/core/Const.kt
```diff
@@ -15,12 +15,10 @@ data class Const<A, out T>(private val value: A) : ConstOf<A, T> {
   fun <U> retag(): Const<A, U> =
     this as Const<A, U>
 
-  @Suppress("UNUSED_PARAMETER")
-  fun <G, U> traverse(GA: Applicative<G>, f: (T) -> Kind<G, U>): Kind<G, Const<A, U>> =
+  fun <G, U> traverse(GA: Applicative<G>): Kind<G, Const<A, U>> =
```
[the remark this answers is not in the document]
aballano: I just reverted the commit, to me Const it's a bit confusing, so better leave it be :D

arrow-core-data/src/main/kotlin/arrow/core/Eval.kt
aballano: This is caused due to a likely false positive of Animal-Sniffer not being able to process this, as soon as I added the inline it started failing here 🤷

nomisRev: Hmm, could it be that Kotlinc is generating code not compatible with Android? This is the only doc I could find on that annotation, but it clearly states that it is used to mark code sites that `knowingly use features that go beyond the minimum requirement`.

https://github.com/mojohaus/animal-sniffer/blob/fdc17216a51a57a408d384e6af12830ff685cb5f/animal-sniffer/src/site/apt/signature-checker.apt#L123

aballano:
> Not 100% sure what is going on there with Animal Sniffer, but we could verify this on Android once merged.
> 
> Should we create a ticket, to make sure we don't forget to test? Guess we need to run on our lowest support API, 21?

I'll test it ASAP

PR #193, opened 2020-07-23
Monoids, Monoids and some semigroups

1Jajen1:
I added some common monoids which are very useful when for interaction with any sort of monoid/semigroup based interface (foldable, reducers, etc).

I also deprecated the number types monoids because there are a lot of different monoids for numbers and by requiring a wrapper we offer a lot more control.

I also made `Endo` use `AndThen` on composition to avoid stackoverflows with it.

Lastly one addition I did not want to make but likely have to is the `Num` typeclass for the `Sum` and `Product` wrappers. I needed access to `*` and `+`, but the kotlin class `Number` does not have them. `Num` is lawless because testing its laws with anything but arbitrary precision types isn't worth it. Any ideas on how to avoid this typeclass?


1Jajen1:
This will also need doc entries and I haven't checked if ank compiles yet, but I am out of time for today :)

1Jajen1:
Oh and give me newtype like deriving, this pr is probably ~300 lines if I could exclude all trivially derivable instances...

1Jajen1:
Oh and for reference all of these conform to the ones from `Data.Semigroup` and `Data.Monoid` from haskells prelude.

https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Semigroup.html
https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Monoid.html

nomisRev:
Oh this is still open as well... 😅 Sorry @1Jajen1!

I went through all this some time ago, but need to make some time to go through it again. It looked great to me, but will review again this week!
I added tagged some other reviewers as well too.

raulraja (approved):

raulraja:
Changes look great but I wish they could be non dependent of core at least until we decide on the strategy regarding effect handlers vs typeclasses and if they can coexist or it doesn't make sense. @1Jajen1 happy to retake that conversation whenever you want ☺️

1Jajen1:
I am actually not sure about this pr as of now. These types are great if we have newtypes and typeclass inference. Otherwise they get a bit annoying to use.

The main use for this is to enrich typeclass (monoid/semigroup) apis like `Foldable`. In haskell they are also used in deriving strategies to create custom instances without writing any code for those. Newtypes, typeclass inference and deriving strategies are easily one of the most powerful combinations that haskell has to offer and unlike other more abstract haskell features this one is 100% desirable everywhere.

I am not sure this should be in core just yet, and I think postponing this until we have a definitive encoding for typeclasses, newtypes and maybe even deriving could be a better idea. That being said, if you guys think this is low enough maintenance to just add, then we could also just merge it and keep it around until it is easier to use.

nomisRev:
> low enough maintenance

&

> +2,142 −23

I am fine postponing this change 😅

raulraja:
Let's postpone then until we figure out the final public apis for the upcoming 1.x

1Jajen1:
> > low enough maintenance
> 
> &
> 
> > +2,142 −23
> 
> I am fine postponing this change

It is 99% redundant code for instances :see_no_evil: 

But yeah I agree, better wait till some better way of encoding these shows up :+1:

PR #194, opened 2020-07-24
Hash typeclass improvements

1Jajen1:
I made a few changes to the `Hash` typeclass to make it both easier to define and better to use.

- Removed the `Eq` superclass from it. It never made sense and it should not have been there in the first place.
- Added `hashWithSalt`. Allows for two nice things: First hashing with salt, but also allows very easy definitions for sum types: `Either<L, R>.hash(HL: Hash<L>, HR: Hash<R>): Int = fold({ l -> HL.run { l.hashWithSalt(0) } }, { r -> HR.run { r.hashWithSalt(1) } })`
- Added `Hashed` datatype. It carriers its precomputed hash and thus provides very fast inequality checks and provides its `Hash` instances with the cached hash.

Overall this makes `Hash` actually somewhat useful over the default `hashCode` now :)


rachelcarmena:
I removed the last merge with `master` branch because errors raised after solving the conflict. In this way, the conflict can be solved again @1Jajen1 :wink:

PR #195, opened 2020-07-24
Remove the all value from Nel

### LINKED ISSUE #190: Nel<T> is not actually a List<T>. Should this be changed?
**What version are you currently using?**
0.10.5

**What would you like to see?**
I'd like to see Nel actually implementing `List<T>`. It annoys me pretty often that I have to use the `all` property and cannot pass Nel to where a list is needed.
Kotlin has great starting point for making custom list implementations, it's the [ `AbstractList<T>` class](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-abstract-list/). Was this considered during implementation? 
I always feel a bit distracted that arrow's Nel uses twice the amount of memory it actually needs: one for the `head` & `rest` and exactly the same amount for the `all`.
Here what it can look like, if we implemented it via `AbstractList`:
```kotlin
class NonEmptyList<T> (val first: T, val rest: List<T>): AbstractList<T>() {
	override val size = 1 + rest.size

	override operator fun get(index: Int): T {
		if (index < 0 || index >= size) throw IndexOutOfBoundsException(index)
		return if (index == 0) first else rest[index - 1]
	}
}
```
It is just as simple as that. And we get alot of functionality for free:
- Reduced memory consumption (compared to the current implementation of Nel),
- equals/hashCode
- toString()
- Iterator
- Indexing
Just about everything that `List<T>` has. And what's more important: the class implements `List<T>` now (and thus, `Collection<T>`, `Iterable<T>`) which allows:
- use it in place where one of those needed
- use all the hell of extension functions provided by the kotlin standard lib.

Of course we have all these features via `all` property. But why not having it on Nel directly, if it is that simple?

1Jajen1:
As #190 described this is a huge waste of memory.

This also makes `NonEmptyList` implement `AbstractList` which opens it up for a bunch of nice extension functions from the stdlibs. I am not sure how much this clashes with our api, but kotlin will choose the most specific type, which should be `NonEmptyList` right?

Closes #190


nomisRev:
Yes, whenever the function is defined directly on the `NonEmptyList` it will prefer that.

The only downside of the extension functions is that it'll turn `NonEmptyList` into `List`,
but that's fine in *some* cases such as `filter` or `mapNotNull` since those are not valid functions for `(Nel) -> Nel`.

arrow-core-data/src/main/kotlin/arrow/core/NonEmptyList.kt
```diff
@@ -194,30 +194,27 @@ typealias Nel<A> = NonEmptyList<A>
...
+) : NonEmptyListOf<A>, AbstractList<A>() {
 
-  constructor(head: A, tail: List<A>) : this(head, tail.toList(), listOf(head) + tail.toList())
-  private constructor(list: List<A>) : this(list[0], list.drop(1), list.toList())
+  private constructor(list: List<A>) : this(list[0], list.drop(1))
 
-  val size: Int =
-    all.size
+  override val size: Int
+    get() = 1 + tail.size
```
nomisRev, line 205:
```suggestion
  override val size: Int =
    1 + tail.size
```

1Jajen1:
> Yes, whenever the function is defined directly on the `NonEmptyList` it will prefer that.

Good that means it won't be a nightmare to use :)

> The only downside of the extension functions is that it'll turn `NonEmptyList` into `List`,
> but that's fine in _some_ cases such as `filter` or `mapNotNull` since those are not valid functions for `(Nel) -> Nel`.

I actually think that it is fine in most if not all cases: If you want to stay with `Nel` you'll have to use our api which is safe. We never have any control over other methods taking `AbstractList` thus we can only assume that to be unsafe anyway.

arrow-core-data/src/main/kotlin/arrow/core/NonEmptyList.kt
```diff
@@ -194,30 +194,27 @@ typealias Nel<A> = NonEmptyList<A>
  *
  */
 @higherkind
-class NonEmptyList<out A> private constructor(
+class NonEmptyList<out A>(
   val head: A,
-  val tail: List<A>,
-  val all: List<A>
-) : NonEmptyListOf<A>, Iterable<A> by all {
```
nomisRev, line 201: We should've already supported all extension functions from the std since we extended `Iterable`.

nomisRev:
> you'll have to use our api which is safe

Except for `get`, but that was also already available through `Iterable` today.

arrow-core-data/src/main/kotlin/arrow/core/NonEmptyList.kt
1Jajen1: Well `AbstractList` has `Iterable` in its hierarchy so it should be more powerful like this.

1Jajen1:
Looks like this breaks the serialization from `arrow-integrations`. Time to learn how those work :see_no_evil:

1Jajen1:
> Looks like this breaks the serialization from `arrow-integrations`. Time to learn how those work

So my guess is that this change, likely introducing `AbstractList` into the hierarchy makes jackson think it can just use the `StdListDeserializer` and be done with it. This is super weird as we have registered our own delegating deserializer and it just gets ignored. I find jackson to be way too weird to actually wrap my head around (like I can read aesons generic derivation code just fine but this reflection witchcraft is just confusing) so take this reasoning with a grain of salt, I basically have no idea if that's correct.

Any ideas on what to do here?

nomisRev:
@1Jajen1 very likely that the reflection machinery picks up the incorrect type and is done with it :/

nomisRev:
I guess `: List<A> by all` would solve the same issue, right? Perhaps with a less ideal encoding?

On the other side, I'm not sure we should ditch this improvement over something weird going on with Jackson & reflection.

nomisRev:
@rachelcarmena @raulraja @aballano @JorgeCastilloPrz

This PR is currently blocked by **arrow-integrations-jackson-module**, which was incompletely contributed and never maintained by anyone. So I'm not sure this should be included in the integration build step.

Additionally, concerning **arrow-integrations-retrofit-adapter** it might be more interesting to offer this module in Arrow Core (or Arrow Fx). Since we're promoting `suspend` over `IO` from `0.11.0`, which retrofit already provides out-of-the-box, we only need support for `MonadError` (Arrow-Core). Which is a small and focused micro-library to connect Arrow Core data types to Retrofit.

Thoughts?

aballano:
> @rachelcarmena @raulraja @aballano @JorgeCastilloPrz
> 
> This PR is currently blocked by **arrow-integrations-jackson-module**, which was incompletely contributed and never maintained by anyone. So I'm not sure this should be included in the integration build step.
> 
> Additionally, concerning **arrow-integrations-retrofit-adapter** it might be more interesting to offer this module in Arrow Core (or Arrow Fx). Since we're promoting `suspend` over `IO` from `0.11.0`, which retrofit already provides out-of-the-box, we only need support for `MonadError` (Arrow-Core). Which is a small and focused micro-library to connect Arrow Core data types to Retrofit.
> 
> Thoughts?

Regarding the **arrow-integrations-jackson-module**, the current issue seems to be the lack of a default constructor for Nel, which I guess is caused by the change in implementation.

Regarding the Integrations I agree that an arrow-fx-coroutines-integrations should be needed sooner or later 👍 

On this topic we currently have this issue (with a correspondent PR) https://github.com/arrow-kt/arrow-integrations/issues/27 that we could adapt/close for this matter.

1Jajen1:
> > @rachelcarmena @raulraja @aballano @JorgeCastilloPrz
> > This PR is currently blocked by **arrow-integrations-jackson-module**, which was incompletely contributed and never maintained by anyone. So I'm not sure this should be included in the integration build step.
> > Additionally, concerning **arrow-integrations-retrofit-adapter** it might be more interesting to offer this module in Arrow Core (or Arrow Fx). Since we're promoting `suspend` over `IO` from `0.11.0`, which retrofit already provides out-of-the-box, we only need support for `MonadError` (Arrow-Core). Which is a small and focused micro-library to connect Arrow Core data types to Retrofit.
> > Thoughts?
> 
> Regarding the **arrow-integrations-jackson-module**, the current issue seems to be the lack of a default constructor for Nel, which I guess is caused by the change in implementation.
> 
> Regarding the Integrations I agree that an arrow-fx-coroutines-integrations should be needed sooner or later 
> 
> On this topic we currently have this issue (with a correspondent PR) [arrow-kt/arrow-integrations#27](https://github.com/arrow-kt/arrow-integrations/issues/27) that we could adapt/close for this matter.

No that's not quite it. It fails to grab the DelegatingDeserializers by type and then tries to instantiate a list. But since we don't have such a constructor it fails with that rather confusing error.

rachelcarmena:
> @rachelcarmena @raulraja @aballano @JorgeCastilloPrz
> 
> This PR is currently blocked by **arrow-integrations-jackson-module**, which was incompletely contributed and never maintained by anyone. So I'm not sure this should be included in the integration build step.
> 
> Additionally, concerning **arrow-integrations-retrofit-adapter** it might be more interesting to offer this module in Arrow Core (or Arrow Fx). Since we're promoting `suspend` over `IO` from `0.11.0`, which retrofit already provides out-of-the-box, we only need support for `MonadError` (Arrow-Core). Which is a small and focused micro-library to connect Arrow Core data types to Retrofit.
> 
> Thoughts?

LGTM. After moving `arrow-integrations-retrofit-adapter` module, just remove `arrow-integrations` from these lists: https://github.com/arrow-kt/arrow/tree/master/lists

nomisRev:
To unblock this PR we need to decide what to do with the module "arrow-integrations-retrofit-adapter".
This module currently contains 2 different integrations actually.

 - An adapter for `Either<ErrorBody, ResponseBody>` by @LordRaydenMK in https://github.com/arrow-kt/arrow-integrations/pull/26
 - `CallK` based on top `Async`, `MonadDefer` & `MonadError` by @leandroBorgesFerreira in https://github.com/arrow-kt/arrow/pull/1037

The first being an Arrow-Core integration since Retrofit already support `suspend` out of the box. This would offer a full integration with Arrow Core & Arrow Fx Coroutines.

The second contribution depends on Arrow Fx's IO, which will be deprecated for `0.11.0`.

So if propose we move the contribution from @LordRaydenMK to a new module in Arrow-Core, then we've officially promoted that module from being an incubating integration to an non-incubating integration and we can exclude `arrow-integrations` from CI due to IO's deprecation.

arrow-core-data/src/main/kotlin/arrow/core/NonEmptyList.kt
raulraja: The problem with AbstractList is that exposes operations that are unsafe in Nel like get.
I think this should just implement Iterable structurally so its used in Kotlin loops etc but does not offer unsafe apis like get. Iterable is weaker and still projects all the collections methods from the std lib to terminal ops based on List. 
For the Kotlin docs, we just need structural extensions not even conforming to Iterable if we didn't want the projections which I believe we do.
> As mentioned before, for iterates through anything that provides an iterator, i.e.
> has a member- or extension-function iterator(), whose return type
> has a member- or extension-function next(), and
> has a member- or extension-function hasNext() that returns Boolean.
> All of these three functions need to be marked as operator.

https://kotlinlang.org/docs/reference/control-flow.html#for-loops

JorgeCastilloPrz:
As @nomisRev said, Retrofit provides support for sync style by `suspend` since quite long ago already, so you don't need any call adapters nor wrappers over your service return types anymore:

```kotlin
Either.catch { service.loadUsers() }
```

That's all the retrofit integration you need for any functional codebase. I don't see a big reason for providing any of the two mentioned points.

nomisRev:
@JorgeCastilloPrz the adapter for either works in such a way that the `Throwable` would be captured inside `suspend` and any `Response` errors are returned inside the left side of an `Either` in a typed way.

Such that you don't have to manually detangle the `Throwable` etc to figure out what went wrong.

From the documentation in the linked PR by @LordRaydenMK:

> * Using [Either] as the return type means that 200 status code and HTTP errors return a value,
 > * other exceptions will throw.

PR #196, opened 2020-07-24
Replace `Option<T>` with `?` in `ListK` (part 1)

tapegram:
Part 1 of providing alternatives for fns that use `Option` for `ListK`.

Ran into https://discuss.kotlinlang.org/t/overloaded-function-with-lambda-parameter-of-different-return-types/6053/2 when creating a nullable version of `filterMap`, so I gave the nullable one a different and less good name :(

Would especial appreciate any feedback on the naming of these functions, since the older `Option` based functions already have the good names and the type system struggles to handle overloading with generics or lambdas without requiring explicit typing by the consumer, so I had to make up new names when required.

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/core/ListK.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/ListK.kt
@@ -283,11 +283,39 @@ data class ListK<out A>(private val list: List<A>) : ListKOf<A>, List<A> by list
   /**
    * Left-padded zipWith.
    */
+  @Deprecated("Deprecated, use `leftPadZip(other: ListK<B>, fab: (A?, B) -> C)` instead", ReplaceWith("leftPadZip(other: ListK<B>, fab: (A?, B) -> C)"))
   fun <B, C> lpadZipWith(
     other: ListK<B>,
     fab: (Option<A>, B) -> C
   ): ListK<C> =
     padZipWith(other) { a: Option<A>, b -> b.map { fab(a, it) } }.filterMap(::identity)
+
+  /**
+   * Returns a [ListK<C>] containing the result of applying some transformation `(A?, B) -> C`
+   * on a zip, excluding all cases where the right value is null.
+   *
+   * Example:
+   * ```kotlin:ank:playground
+   * import arrow.core.*
+   *
+   * //sampleStart
+   * val left = listOf(1, 2).k().leftPadZip(listOf("a").k()) { l, r -> l toT r }.k()      // Result: ListK(Tuple2(1, "a"))
+   * val right = listOf(1).k().leftPadZip(listOf("a", "b").k()) { l, r -> l toT r }.k()   // Result: ListK(Tuple2(1, "a"), Tuple2(null, "b"))
+   * val both = listOf(1, 2).k().leftPadZip(listOf("a", "b").k()) { l, r -> l toT r }.k() // Result: ListK(Tuple2(1, "a"), Tuple2(2, "b"))
+   * //sampleEnd
+   *
+   * fun main() {
+   *   println("left = $left")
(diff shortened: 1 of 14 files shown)


nomisRev:
Oh, that sucks :( This should be fixable with `@BuilderInference`, but that's still experimemtal isn't it?

nomisRev:
Hey @tapegram,

Since Arrow is an OSS project we use KtLint to check to format and to automatically format. It seems that CI failed due to a formatting issue.

Running `.gradlew ktlintFormat` should correctly format Arrow, checking that in and pushing should make CI green so we can make this PR ready for review 🙌

arrow-core-data/src/main/kotlin/arrow/core/ListK.kt
```diff
@@ -167,14 +167,41 @@ data class ListK<out A>(private val list: List<A>) : ListKOf<A>, List<A> by list
...
+   *   }
+   * } // Result: listOf("2").k()
+   * //sampleEnd
+   *
+   * fun main() {
+   *   println("evenStrings = $evenStrings")
+   * }
+   * ```
+   */
+  fun <B> filterNullMap(f: (A) -> B?): ListK<B> =
```
nomisRev, line 196: Should we do? We're also trying to algining more to the names familiar in Kotlin from the std. IIRC there is 2 `Iterable` extensions for the signature in the std. `filterMap` and `mapNotNull`?

```
@JvmName("mapNotNull")
fun <B> filterMap(f: (A) -> B?): ListK<B>
```

1Jajen1: I think as long as the stdlib ones don't have any obvious drawbacks we shouldn't have duplicates names and signatures.

arrow-core-data/src/main/kotlin/arrow/core/ListK.kt
```diff
@@ -185,23 +212,110 @@ data class ListK<out A>(private val list: List<A>) : ListKOf<A>, List<A> by list
...
+   * //sampleEnd
+   *
+   * fun main() {
+   *   println("padRight = $padRight")
+   *   println("padLeft = $padLeft")
+   *   println("noPadding = $noPadding")
+   * }
+   * ```
+   */
+  fun <B> padZipWithNull(
```
1Jajen1, line 236: Can't we have them use the same name and disambiguate for the jvm using jvmName or something?

arrow-core-data/src/main/kotlin/arrow/core/ListK.kt
nomisRev: In that case I think we should name it [`mapNotNull` as Kotlin's Std](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/map-not-null.html)? 

We can re-introduce `filterMap` if we want later on.

arrow-core-data/src/main/kotlin/arrow/core/ListK.kt
```diff
@@ -167,14 +167,41 @@ data class ListK<out A>(private val list: List<A>) : ListKOf<A>, List<A> by list
...
+   * ```kotlin:ank:playground
+   * import arrow.core.*
+   *
+   * //sampleStart
+   * val evenStrings = listOf(1, 2).k().filterNullMap {
+   *   when (it % 2 == 0) {
+   *     true -> it.toString()
+   *     else -> null
+   *   }
+   * } // Result: listOf("2").k()
```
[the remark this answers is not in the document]
nomisRev: This is actually not the case, that's only true for non-playground snippets and the output only appears on the website.

nomisRev:
Thanks so much @tapegram for the contribution!

PR #197, opened 2020-07-24
Rework Order to use Ordering and test instance laws

1Jajen1:
Two changes worth noting:
- Orders main combinator now returns `Ordering`. This has a few advantages: It is much easier to reason about than integers and second it offers a monoid instance so we can combine the result of `compare` to obtain a new `Ordering`. This allows nice and easy definitions for types with multiple values.
- Added laws and tests for the `Order` typeclass. It was previously lawless.

### DIFF
--- a/arrow-core/src/main/kotlin/arrow/core/extensions/boolean.kt
+++ b/arrow-core/src/main/kotlin/arrow/core/extensions/boolean.kt
@@ -3,6 +3,8 @@ package arrow.core.extensions
 import arrow.typeclasses.Eq
 import arrow.typeclasses.Hash
 import arrow.typeclasses.Monoid
+import arrow.typeclasses.Order
+import arrow.typeclasses.Ordering
 import arrow.typeclasses.Show
 
 interface BooleanShow : Show<Boolean> {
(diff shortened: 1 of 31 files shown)


arrow-core/src/main/kotlin/arrow/core/extensions/boolean.kt
```diff
@@ -14,6 +16,11 @@ interface BooleanEq : Eq<Boolean> {
   override fun Boolean.eqv(b: Boolean): Boolean = this == b
 }
 
+interface BooleanOrder : Order<Boolean> {
+  override fun Boolean.compare(b: Boolean): Ordering = Ordering.fromInt(this.compareTo(b))
```
nomisRev, line 20: Nit: impls on new lines

nomisRev:
Thanks @1Jajen1! I think this was one of my early contributions to Arrow 😅
Using `Int` over `Ordering` was definitely the wrong call.

1Jajen1:
The build fails with a bug in kapt by the looks of it. Locally compiles fine. This looks weird as if it has a cache issue because it seems related to `Eq` being a superclass of `Order` which was removed in this pr

1Jajen1:
> The build fails with a bug in kapt by the looks of it. Locally compiles fine.

Got this fixed. My guess is that github actions run on a machine that considers the paths: "x/EQ.java" and "x/Eq.java" to be the same whereas my machine does not. Only way I can explain why this failed in actions only :sweat_smile: 

Either way I moved the `Ordering` to the datatypes where it belonged in the first place.

1Jajen1:
> arrow.fx.QueueTest > BoundedQueue - peekAll reads all values, including outstanding offers FAILED

I guess that is a flaky test?

nomisRev:
@1Jajen1 yes, it's flaky due to CI imprecision on timing :/

rachelcarmena:
> My guess is that github actions run on a machine that considers the paths: "x/EQ.java" and "x/Eq.java" to be the same whereas my machine does not. 

Right @1Jajen1 , it runs on a macOS environment with case insensitive filesystem :raised_hands:

PR #221, opened 2020-08-24
Configuration: improve plugin management

rachelcarmena:


.editorconfig
```diff
@@ -3,4 +3,4 @@
 indent_size=2
 insert_final_newline=true
 max_line_length=off
-disabled_rules=import-ordering,no-unit-return
\ No newline at end of file
+disabled_rules=import-ordering,no-unit-return,curly-spacing
```
aballano, line 6: was this added on purpose? AFAIK ktlint shouldn't affect non-kotlin files

rachelcarmena: Righ, thanks for the review!! When re-organizing the configuration, the ktlint version was considered and a lot of errors raised with functions following this format:

```
<signature> = 
  { ... }
```

I updated this configuration for all the repositories except `arrow-check` (I'll create an issue afterwards).

aballano (approved):

PR #222, opened 2020-08-25
Updating `MapK` to have `Option` alternatives

tapegram:
Update a few functions that use Option to use nullables instead.
Deprecate two functions and provide alternatives.

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
@@ -97,7 +97,6 @@ fun <K, A> Pair<K, A>?.asIterable(): Iterable<Pair<K, A>> =
     else -> listOf(this)
   }
 
-
 fun <K, A, B> Map<K, A>.foldRight(b: Map<K, B>, f: (Map.Entry<K, A>, Map<K, B>) -> Map<K, B>): Map<K, B> =
   this.entries.reversed().k().foldLeft(b) { x, y: Map.Entry<K, A> -> f(y, x) }
 
--- a/arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
@@ -65,15 +65,12 @@ data class MapK<K, out A>(private val map: Map<K, A>) : MapKOf<K, A>, Map<K, A> 
 
 fun <K, A> Map<K, A>.k(): MapK<K, A> = MapK(this)
 
-@Deprecated("Deprecated, use nullable instead", ReplaceWith("Tuple2<k, A>?.k()"))
+@Deprecated("Deprecated, use nullable instead", ReplaceWith("Tuple2<K, A>>?.let { ... }"))
 fun <K, A> Option<Tuple2<K, A>>.k(): MapK<K, A> =
   when (this) {
     is Some -> mapOf(this.t).k()
     is None -> emptyMap<K, A>().k()
   }
-
-fun <K, A> Tuple2<K, A>?.k(): MapK<K, A> =
-  this?.let { mapOf(this).k() } ?: emptyMap<K, A>().k()
 
 fun <K, V, G> MapKOf<K, Kind<G, V>>.sequence(GA: Applicative<G>): Kind<G, MapK<K, V>> =
   fix().traverse(GA, ::identity)
@@ -91,7 +88,7 @@ fun <K, A, B> Map<K, A>.foldLeft(b: Map<K, B>, f: (Map<K, B>, Map.Entry<K, A>) -
   return result
 }
 
-fun <K, A> Pair<K, A>?.asIterable(): Iterable<Pair<K, A>> =
+internal fun <K, A> Pair<K, A>?.asIterable(): Iterable<Pair<K, A>> =
   when (this) {
     null -> emptyList()
     else -> listOf(this)
--- a/arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
@@ -65,6 +65,7 @@ data class MapK<K, out A>(private val map: Map<K, A>) : MapKOf<K, A>, Map<K, A> 
 
 fun <K, A> Map<K, A>.k(): MapK<K, A> = MapK(this)
 
+@Deprecated("Deprecated, use nullable instead", ReplaceWith("Tuple2<k, A>?.k()"))
 fun <K, A> Option<Tuple2<K, A>>.k(): MapK<K, A> =
   when (this) {
     is Some -> mapOf(this.t).k()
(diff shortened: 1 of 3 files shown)


arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
```diff
@@ -13,7 +13,7 @@ data class MapK<K, out A>(private val map: Map<K, A>) : MapKOf<K, A>, Map<K, A>
   fun <B, Z> map2(fb: MapK<K, B>, f: (A, B) -> Z): MapK<K, Z> =
     if (fb.isEmpty()) emptyMap<K, Z>().k()
     else this.map.flatMap { (k, a) ->
-      fb.getOption(k).map { Tuple2(k, f(a, it)) }.k().asIterable()
+      fb[k]?.let { Tuple2(k, f(a, it)) }.k().asIterable()
```
nomisRev, line 16: Do we need the conversion to `KList` with `k()` in between?

nomisRev (approved):
Awesome! Thanks @tapegram.
Small nit, I think we can remove some allocations :)

nomisRev (approved):
Thanks @tapegram! This will safe us some allocations in the implementations 🎉 

Seems you still need to run `ktlintFormat`, I tried to do it for you but couldn't push to this branch.

arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
```diff
@@ -13,8 +13,8 @@ data class MapK<K, out A>(private val map: Map<K, A>) : MapKOf<K, A>, Map<K, A>
   fun <B, Z> map2(fb: MapK<K, B>, f: (A, B) -> Z): MapK<K, Z> =
     if (fb.isEmpty()) emptyMap<K, Z>().k()
     else this.map.flatMap { (k, a) ->
-      fb.getOption(k).map { Tuple2(k, f(a, it)) }.k().asIterable()
-    }.k()
+      fb[k]?.let { Pair(k, f(a, it)) }.asIterable()
```
aballano, line 16: nit: not a must, but I'd keep the Tuple2 here and below if possible for consistency

```suggestion
      fb[k]?.let { Tuple2(k, f(a, it)) }.asIterable()
```

arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
```diff
@@ -25,14 +25,14 @@ data class MapK<K, out A>(private val map: Map<K, A>) : MapKOf<K, A>, Map<K, A>
 
   fun <B, Z> ap2(f: MapK<K, (A, B) -> Z>, fb: MapK<K, B>): Map<K, Z> =
     f.map.flatMap { (k, f) ->
-      this.flatMap { a -> fb.flatMap { b -> mapOf(Tuple2(k, f(a, b))).k() } }
-        .getOption(k).map { Tuple2(k, it) }.k().asIterable()
-    }.k()
+      this.flatMap { a -> fb.flatMap { b -> mapOf(Tuple2(k, f(a, b))).k() } }[k]
+        ?.let { Pair(k, it) }.asIterable()
```
aballano, line 29:
```suggestion
        ?.let { Tuple2(k, it) }.asIterable()
```

arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
```diff
@@ -25,14 +25,14 @@ data class MapK<K, out A>(private val map: Map<K, A>) : MapKOf<K, A>, Map<K, A>
...
+    }.toMap().k()
 
   fun <B> flatMap(f: (A) -> MapK<K, B>): MapK<K, B> =
     this.map.flatMap { (k, v) ->
-      f(v).getOption(k).map { Tuple2(k, it) }.k().asIterable()
-    }.k()
+      f(v)[k]?.let { Pair(k, it) }.asIterable()
```
aballano, line 34:
```suggestion
      f(v)[k]?.let { Tuple2(k, it) }.asIterable()
```

arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
```diff
@@ -86,6 +91,12 @@ fun <K, A, B> Map<K, A>.foldLeft(b: Map<K, B>, f: (Map<K, B>, Map.Entry<K, A>) -
   return result
 }
 
+fun <K, A> Pair<K, A>?.asIterable(): Iterable<Pair<K, A>> =
```
aballano, line 94:
```suggestion
fun <K, A> Tuple2<K, A>?.asIterable(): Iterable<Tuple2<K, A>> =
```

aballano (approved):
Neat 👌

arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
```diff
@@ -65,17 +65,22 @@ data class MapK<K, out A>(private val map: Map<K, A>) : MapKOf<K, A>, Map<K, A>
...
+@Deprecated("Deprecated, use nullable instead", ReplaceWith("Tuple2<k, A>?.k()"))
 fun <K, A> Option<Tuple2<K, A>>.k(): MapK<K, A> =
   when (this) {
     is Some -> mapOf(this.t).k()
     is None -> emptyMap<K, A>().k()
   }
 
+fun <K, A> Tuple2<K, A>?.k(): MapK<K, A> =
+  this?.let { mapOf(this).k() } ?: emptyMap<K, A>().k()
+
```
aballano, line 77: I believe this is not used, is it?

```suggestion
```

arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
tapegram: @nomisRev suggested we switch to pair for this one. The impression I got is that Pair is slightly more efficient than Tuple2 in this case.

arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
tapegram: Copying message from other comment: @nomisRev suggested we switch to pair for this one. The impression I got is that Pair is slightly more efficient than Tuple2 in this case.

arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
tapegram: It's a replacement for `Option<Tuple2<K, A>>.k()` above.

arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
nomisRev: Using `Pair` allows using Kotlin's standard library functions which we'd otherwise still have to provide. I thought we'd prefer Kotlin Std.

arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
aballano: I see, but I doubt this function should be exposed to the public (unless @nomisRev thinks otherwise) because it makes no sense name-wise and it should be internal. Now being internal, if we don't use it, then it can be completely removed I believe

arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
aballano, line 94: This should be internal as well I believe

arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
tapegram: ah you are right, I will update it to have a better name and probably to be private.

arrow-core/src/main/kotlin/arrow/core/extensions/mapk.kt
```diff
@@ -42,7 +42,6 @@ import arrow.typeclasses.Traverse
 import arrow.typeclasses.Zip
 import arrow.typeclasses.Unalign
 import arrow.typeclasses.Unzip
-import arrow.typeclasses.hashWithSalt
```
tapegram, line 45: Think this got picked up as unused by ktlintformat

aballano (approved):
Good job @tapegram, thank you!

PR #224, opened 2020-08-25
Configuration: remove transitive dependency and update configurations

rachelcarmena:
This dependency is been added to all the Arrow Core modules except `arrow-meta`:

```
    <dependency>
      <groupId>org.codehaus.mojo</groupId>
      <artifactId>animal-sniffer-annotations</artifactId>
      <version>1.19</version>
      <scope>runtime</scope>
    </dependency>
```

This PR removes that transitive dependency and updates the configuration for dependencies because of https://github.com/arrow-kt/arrow/pull/2206


arrow-meta/build.gradle
```diff
@@ -10,28 +10,28 @@ apply from: "$SUB_PROJECT"
 apply from: "$DOC_CREATION"
 
 dependencies {
-    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$KOTLIN_VERSION"
-    compile project(':arrow-annotations')
-    compile "me.eugeniomarletti.kotlin.metadata:kotlin-metadata:$KOTLIN_METADATA_VERSION"
-    compile("com.squareup:kotlinpoet:$KOTLIN_POET_VERSION") {
+    api "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$KOTLIN_VERSION"
```
aballano, line 13: I'm not sure this should be exposed as it imposes a specific version + jdk to other places, for example in our own repo we have jdk7, jdk8 and maybe some plain jdk declarations that would be duplicated

```suggestion
    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$KOTLIN_VERSION"
```

arrow-meta/build.gradle
```diff
@@ -10,28 +10,28 @@ apply from: "$SUB_PROJECT"
...
+    api project(':arrow-annotations')
+    api "me.eugeniomarletti.kotlin.metadata:kotlin-metadata:$KOTLIN_METADATA_VERSION"
+    api("com.squareup:kotlinpoet:$KOTLIN_POET_VERSION") {
```
aballano, line 16: Those should stay within arrow-meta only AFAIK

```suggestion
    implementation "me.eugeniomarletti.kotlin.metadata:kotlin-metadata:$KOTLIN_METADATA_VERSION"
    implementation("com.squareup:kotlinpoet:$KOTLIN_POET_VERSION") {
```

arrow-meta/build.gradle
```diff
@@ -10,28 +10,28 @@ apply from: "$SUB_PROJECT"
...
         exclude group: 'org.jetbrains.kotlin', module: 'kotlin-reflect'
     }
-    compile "org.jetbrains.kotlin:kotlin-reflect:$KOTLIN_VERSION"
+    api "org.jetbrains.kotlin:kotlin-reflect:$KOTLIN_VERSION"
```
aballano, line 19: Same

```suggestion
    implementation "org.jetbrains.kotlin:kotlin-reflect:$KOTLIN_VERSION"
```

arrow-core/build.gradle
```diff
@@ -8,13 +8,13 @@ apply from: "$SUB_PROJECT"
 apply from: "$DOC_CREATION"
 
 dependencies {
-    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$KOTLIN_VERSION"
-    compile project(":arrow-annotations")
-    compile project(":arrow-core-data")
+    api "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$KOTLIN_VERSION"
```
aballano, line 11:
```suggestion
    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$KOTLIN_VERSION"
```

arrow-core/build.gradle
```diff
@@ -8,13 +8,13 @@ apply from: "$SUB_PROJECT"
...
+    api project(":arrow-annotations")
```
aballano, line 12: I believe it should suffice by exposing this from arrow-meta only as it's the module taking care of them + doing annotation processing

arrow-core/build.gradle
```diff
@@ -8,13 +8,13 @@ apply from: "$SUB_PROJECT"
...
+    api project(":arrow-core-data")
```
aballano, line 13: This might come handy to save the users from self-importing this themselves, although I'm wondering if [this guide in the arrow readme](https://github.com/arrow-kt/arrow/blame/master/README.md#L105) was then accurate to the previous state 🤔  

I think we should check that document and validate if it's correct (for example arrow-optics doesn't seem to expose core, yet core doesn't appear in the readme for  optics)

arrow-core-test/build.gradle
```diff
@@ -7,11 +7,11 @@ plugins {
...
 
 dependencies {
-    compile project(":arrow-core")
-    compile project(":arrow-core-data")
-    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$KOTLIN_VERSION"
-    compile "org.jetbrains.kotlinx:kotlinx-coroutines-core:$KOTLINX_COROUTINES_VERSION"
-    compile "io.kotlintest:kotlintest-runner-junit5:$KOTLIN_TEST_VERSION", excludeArrow
-    testRuntime "org.junit.vintage:junit-vintage-engine:$JUNIT_VINTAGE_VERSION"
+    api project(":arrow-core")
+    api project(":arrow-core-data")
```
aballano, line 11: I don't think we need to expose those really

```suggestion
    implementation project(":arrow-core")
    implementation project(":arrow-core-data")
```

arrow-core-test/build.gradle
```diff
@@ -7,11 +7,11 @@ plugins {
...
+    api "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$KOTLIN_VERSION"
```
aballano, line 12:
```suggestion
    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$KOTLIN_VERSION"
```

arrow-core-test/build.gradle
```diff
@@ -7,11 +7,11 @@ plugins {
...
+    api "org.jetbrains.kotlinx:kotlinx-coroutines-core:$KOTLINX_COROUTINES_VERSION"
+    api "io.kotlintest:kotlintest-runner-junit5:$KOTLIN_TEST_VERSION", excludeArrow
```
aballano, line 14: I think those 2 are fine to be exported since they're only test-visible, although we should check out any lib importing core-test to potentially remove the dependency from there

arrow-core-retrofit/build.gradle
```diff
@@ -7,12 +7,12 @@ apply from: "$SUB_PROJECT"
...
-    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$KOTLIN_VERSION"
-    compile "com.squareup.retrofit2:retrofit:$RETROFIT_VERSION"
-    testCompile project(":arrow-core-test")
-    testRuntime "org.junit.vintage:junit-vintage-engine:$JUNIT_VINTAGE_VERSION"
-    testCompile "io.kotlintest:kotlintest-runner-junit5:$KOTLIN_TEST_VERSION", excludeArrow
-    testCompile "com.squareup.retrofit2:converter-gson:$RETROFIT_VERSION"
-    testCompile "com.squareup.okhttp3:mockwebserver:$MOCKWEBSERVER_VERSION"
+    api project(":arrow-core-data")
+    api "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$KOTLIN_VERSION"
+    api "com.squareup.retrofit2:retrofit:$RETROFIT_VERSION"
```
aballano, line 12: core-data should be included in core already; regarding retrofit, it's similar as for the kotlin-stdlib, I don't think we should impose a specific version to the users

```suggestion
    implementation project(":arrow-core-data")
    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$KOTLIN_VERSION"
    implementation "com.squareup.retrofit2:retrofit:$RETROFIT_VERSION"
```

arrow-core-data/build.gradle
```diff
@@ -10,15 +10,15 @@ apply from: "$SUB_PROJECT"
 apply from: "$DOC_CREATION"
 
 dependencies {
-    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$KOTLIN_VERSION"
-    compile project(":arrow-annotations")
+    api "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$KOTLIN_VERSION"
+    api project(":arrow-annotations")
```
aballano, line 14:
```suggestion
    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$KOTLIN_VERSION"
    implementation project(":arrow-annotations")
```

arrow-annotations/build.gradle
```diff
@@ -7,6 +7,6 @@ apply from: "$SUB_PROJECT"
 apply from: "$DOC_CREATION"
 
 dependencies {
-    compile "org.jetbrains.kotlin:kotlin-stdlib:$KOTLIN_VERSION"
-    compile "io.kindedj:kindedj:$KINDEDJ_VERSION"
+    api "org.jetbrains.kotlin:kotlin-stdlib:$KOTLIN_VERSION"
+    api "io.kindedj:kindedj:$KINDEDJ_VERSION"
```
aballano, line 11:
```suggestion
    implementation "org.jetbrains.kotlin:kotlin-stdlib:$KOTLIN_VERSION"
    implementation "io.kindedj:kindedj:$KINDEDJ_VERSION"
```

aballano (approved):
Looking good! I think there's some cleanup we might be able to do after this to remove duplicated implementations from other modules 👌

arrow-meta/build.gradle
rachelcarmena: I guess they are necessary to use `arrow-meta`

rachelcarmena:
Thanks for the review @aballano I couldn't apply all the suggestions because those changes imply updating the dependencies. Let's take a look at https://github.com/arrow-kt/arrow/issues/2207

rachelcarmena:
@aballano final result of the differences with this PR. These dependencies aren't exported to consumers now:

**arrow-core-retrofit:**

* arrow-core-data
* kotlin-stdlib-jdk7
* animal-sniffer-annotations
* com.squareup.retrofit2

**arrow-annotations:**

* animal-sniffer-annotations

**arrow-core:**

* kotlin-stdlib-jdk7
* arrow-annotations
* animal-sniffer-annotations 

**arrow-core-data:**

* kotlin-stdlib-jdk7
* animal-sniffer-annotations

**arrow-core-test:**

* kotlin-stdlib-jdk7
* animal-sniffer-annotations

**arrow-meta:**

* kotlin-stdlib-jdk7

**arrow-syntax:**

* kotlin-stdlib-jdk7
* animal-sniffer-annotations

PR #226, opened 2020-08-26
Delimited continuations

1Jajen1:
The purpose of this pr is to add the new module, which @raulraja @nomisRev and me have been working on lately, which implements delimited continuations in kotlin over suspend. (Or something very close to delimited continuations^^)

These implementations are currently split across capabilities (Nesting scopes, multishot) to provide an overview how each feature works. Ideally we'd have Nested scopes and multishot support but that is kind of complicated and thus not included.

The purpose of this pr is (if I understood @raulraja correctly :see_no_evil:) simply adding the module (and api documentation which I will add over the next days) so that further prs can change our existing fx implementations to use delimited continuations in order to slowly chop away the hacky fx code. This should not yet replace fx for nondet carriers (lists/sequences) because I do not consider the multishot included here to be ready. I will create a detailed document/issue for discussion there after this pr is merged.

At some point we also need to have a discussion between folding/binding monadic code versus providing effects directly but thats a different story altogether.

Unless @raulraja or @nomisRev have something to add I'll mark this pr as ready after adding some documentation to each implementation.

Oh and before I forget: I am pretty sure this version does not yet allow interleaving of arbitrary suspend code, so that is also something to look at.

When this pr is merged a few things should happen:
- replace non-multishot fx with a delimited control implementation (this makes them multiplatform viable and likely faster)
- discuss how multishot support should look in arrow (there are a few strategies)
- testing and benchmarking should be set up to see how it performs


arrow-continuations/build.gradle
```diff
@@ -0,0 +1,23 @@
+plugins {
+    id "maven-publish"
+    id "base"
+    id "org.jetbrains.kotlin.jvm"
+    id "org.jetbrains.kotlin.kapt"
+    id "org.jetbrains.dokka"
+    id "org.jlleitschuh.gradle.ktlint"
+    id "ru.vyarus.animalsniffer"
+}
```
rachelcarmena, line 9:
```suggestion
plugins {
    id "org.jetbrains.kotlin.jvm"
    id "org.jetbrains.kotlin.kapt"
    id "org.jlleitschuh.gradle.ktlint"
}
```

arrow-continuations/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+
+apply from: "$SUB_PROJECT"
+apply from: "$DOC_CREATION"
+apply plugin: 'kotlinx-atomicfu'
+
+dependencies {
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$KOTLIN_VERSION"
```
rachelcarmena, line 16:
```suggestion
    compileOnly "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$KOTLIN_VERSION"
```

arrow-continuations/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+    compile project(":arrow-annotations")
+    compile project(":arrow-core")
```
rachelcarmena, line 18: `compileOnly` ?

arrow-continuations/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+    kapt project(":arrow-meta")
+    testRuntime "org.junit.vintage:junit-vintage-engine:$JUNIT_VINTAGE_VERSION"
+    testCompile "io.kotlintest:kotlintest-runner-junit5:$KOTLIN_TEST_VERSION", excludeArrow
+    testCompile project(":arrow-core-test")
```
rachelcarmena, line 22:
```suggestion
    testRuntimeOnly "org.junit.vintage:junit-vintage-engine:$JUNIT_VINTAGE_VERSION"
    testCompileOnly "io.kotlintest:kotlintest-runner-junit5:$KOTLIN_TEST_VERSION", excludeArrow
    testCompileOnly project(":arrow-core-test")
```

arrow-continuations/build.gradle
1Jajen1: I am pretty sure most if not all deps for this module can actually be dropped, will check this tomorrow :+1:

raulraja: This code actually should live inside core because all the Fx blocks there need to be refactored. We initially moved it to its own module to avoid compiling all of core while developing. Unless the fx blocks are reachable from this module we should probably move it to core. @1Jajen1 as you mentioned we need to explore handlers vs instances and related findings before 1.0.

1Jajen1: We can offer continuations as a single module and expose fx constructors from core by adding this as a dep right? 
Continuations have 0 dependencies on core atm, so anyone looking to use them without all the stuff from core can do so.

Not sure what the right call here is, I can copy stuff over either way :+1:

raulraja: If that is the case then perfect I thought we depended in core for some reason. Awesome!

raulraja (approved):
Thank you @1Jajen1 @nomisRev for this journey on continuations and helping make arrow more accessible!

rachelcarmena (approved):
Awesome @1Jajen1 @raulraja @nomisRev Looking forward the next steps!! :raised_hands:

raulraja:
thank you @rachelcarmena for helping out with the merge 🙌

PR #231, opened 2020-08-31
Upgrade to Kotlin 1.4.0

rachelcarmena:
# Changes

* Standard library is added automatically with the plugin now
* Re-order dependencies
* Add `kotlin-reflect` dependency to fix dependency versions
* Use of the new `kotlin-metadata`
* Change the directory for the API doc pieces used to generate the source code (`user.dir` was the current workspace instead of the Kotlin daemon directory; a common directory)
* Fix wrong uses of `contains` because of different types.
* Other fixes because of using a new `kotlin-metadata`


danimontoya (approved):
thanks @rachelcarmena 🚀

nomisRev (approved):

PR #234, opened 2020-09-04
Add foldMapA to Foldable (perhaps by adding lift to Monoid in Applicative)?

m50d:
Hi,

Would it be possible to add a `foldMapA` method to `Foldable`? I'd expect to be able to `foldMap` with a non-monadic applicative.
I've coded up a draft implementation - happy to rework or discard as appropriate (since it's a method on an existing typeclass I couldn't see how to implement it in arrow-incubator - sorry if I've misunderstood, and happy to redo it the right way).
A method for lifting a `Monoid` into an `Applicative` is something I'd also expect to see (and need to use to implement `foldMapA`) - if I've missed an existing implementation then please do point me there.

Thanks,
Mickey

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/typeclasses/Applicative.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/typeclasses/Applicative.kt
@@ -26,4 +26,9 @@ interface Applicative<F> : Apply<F> {
   fun <A> Kind<F, A>.replicate(n: Int, MA: Monoid<A>): Kind<F, A> =
     if (n <= 0) just(MA.empty())
     else mapN(this@replicate, replicate(n - 1, MA)) { (a, xs) -> MA.run { a + xs } }
+
+  fun <A> Monoid<A>.lift(): Monoid<Kind<F, A>> = object : Monoid<Kind<F, A>> {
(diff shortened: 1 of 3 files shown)


nomisRev (approved):
Hey Mickey,

Thanks for your contribution. You're correct this couldn't go in arrow-incubator in a meaningful way, that is meant for new data types or new modules.

Small syntactical Nit:
I think this API would be more clear, and more discoverable as an extension on `Monoid<A>` instead of being an API inside the `Applicative` syntax.

That would allow: `Int.monoid().lift(Either.applicative())` instead of `Either.applicative().run { Int.monoid().lift() }`.

Sorry for the late reply. I've tagged some others for reviewing as well ;)

nomisRev:
PS: You can also always poke in #arrow-contributors on Kotlin Slack for a reviewing or discussing a PR! ;)

m50d:
Thanks! I've moved the `lift` method as you suggest, I agree it's clearer that way.

1Jajen1 (approved):
Looks good to me :+1: 

This pattern is usually done by using `getAp . foldMap Ap` in the haskell world but since we have neither good newtypes nor typeclass resolution this may be easier to use.

PR #188, opened 2020-07-15
Bump kotest

aballano:

PR #237, opened 2020-09-12
Added a resolve function to Either.

joramnv:
Same implementation as the handle function in https://github.com/sparetimedevs/pofpaf/blob/master/src/main/kotlin/com/sparetimedevs/pofpaf/handler/Handler.kt.

I did rename the function to `resolve`, because I like it better than `handle` in combination with `Either`; `Either.resolve`. But I am open to naming suggestions.

Concerning the tests. I wrote the original tests using Kotest. I quickly ported the tests to KotlinTest and they are not optimal. I did see that there is an open PR to upgrade the project to Kotest (https://github.com/arrow-kt/arrow-core/pull/188). If that is merged soon, there won't be a need for the `runBlocking` workaround.

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
@@ -931,55 +931,26 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
      * The resolve function can resolve any suspended function that yields an Either into one type of value.
      *
      * @param f the function that needs to be resolved.
-     * @param ifSuccess the function to apply if [f] yields a success of type [A].
-     * @param ifDomainError the function to apply if [f] yields a domain error of type [E].
-     * @param ifSystemFailure the function to apply if [f] throws a [Throwable].
-     * @param ifHandlingCaseThrows the function to apply if the [ifSuccess], [ifDomainError] or [ifSystemFailure] function throw a [Throwable].
-     * Throwing any [Throwable] in the [ifHandlingCaseThrows] function will render the [resolve] function nondeterministic.
-     * @param ifUnrecoverableState the function to apply if [resolve] is in an unrecoverable state.
+     * @param success the function to apply if [f] yields a success of type [A].
+     * @param error the function to apply if [f] yields a domain error of type [E].
+     * @param throwable the function to apply if [f] throws a [Throwable].
+     * @param handlingCaseThrows the function to apply if the [success], [error] or [throwable] function throw a [Throwable].
+     * Throwing any [Throwable] in the [handlingCaseThrows] function will render the [resolve] function nondeterministic.
+     * @param unrecoverableState the function to apply if [resolve] is in an unrecoverable state.
      * @return the result of applying the [resolve] function.
      */
     suspend fun <E, A, B> resolve(
       f: suspend () -> Either<E, A>,
-      ifSuccess: suspend (a: A) -> Either<Throwable, B>,
-      ifDomainError: suspend (e: E) -> Either<Throwable, B>,
-      ifSystemFailure: suspend (throwable: Throwable) -> Either<Throwable, B>,
-      ifHandlingCaseThrows: suspend (throwable: Throwable) -> Either<Throwable, B> = ifSystemFailure,
-      ifUnrecoverableState: suspend (throwable: Throwable) -> Either<Throwable, Unit> = { Unit.right() }
+      success: suspend (a: A) -> Either<Throwable, B>,
+      error: suspend (e: E) -> Either<Throwable, B>,
+      throwable: suspend (throwable: Throwable) -> Either<Throwable, B>,
+      handlingCaseThrows: suspend (throwable: Throwable) -> Either<Throwable, B> = throwable,
+      unrecoverableState: suspend (throwable: Throwable) -> Either<Throwable, Unit> = { Unit.right() }
     ): B =
       catch { f() }
-        .fold(
-          { throwable: Throwable ->
-            handleItSafely { ifSystemFailure(throwable) }
-          },
-          {
-            it.fold(
-              { e: E ->
-                handleItSafely { ifDomainError(e) }
-              },
-              { a: A ->
-                handleItSafely { ifSuccess(a) }
-              }
-            )
-          }
-        )
-        .fold(
-          { throwable: Throwable ->
-            ifHandlingCaseThrows(throwable)
-          },
-          { b: B ->
-            b.right()
-          }
-        )
-        .fold(
-          { throwable: Throwable ->
-            ifUnrecoverableState(throwable)
-            throw throwable
-          },
-          { b: B ->
-            b
-          }
-        )
+        .fold({ t: Throwable -> handleItSafely { throwable(t) } }, { it.fold({ e: E -> handleItSafely { error(e) } }, { a: A -> handleItSafely { success(a) } }) })
+        .fold({ t: Throwable -> handlingCaseThrows(t) }, { b: B -> b.right() })
+        .fold({ t: Throwable -> unrecoverableState(t); throw t }, { b: B -> b })
   }
 }
 
@@ -1179,13 +1150,5 @@ inline fun <A, B> EitherOf<A, B>.handleErrorWith(f: (A) -> EitherOf<A, B>): Eith
     }
   }
 
-internal suspend fun <A> handleItSafely(f: suspend () -> Either<Throwable, A>): Either<Throwable, A> =
-  Either.catch { f() }
-    .fold(
-      {
-        it.left()
-      },
-      {
-        it
-      }
-    )
+private suspend fun <A> handleItSafely(f: suspend () -> Either<Throwable, A>): Either<Throwable, A> =
+  Either.catch { f() }.fold({ it.left() }, { it })
--- a/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
@@ -926,6 +926,60 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
       } catch (t: Throwable) {
         fe(t.nonFatalOrThrow()).left()
       }
+
+    /**
+     * The resolve function can resolve any suspended function that yields an Either into one type of value.
+     *
+     * @param f the function that needs to be resolved.
+     * @param ifSuccess the function to apply if [f] yields a success of type [A].
+     * @param ifDomainError the function to apply if [f] yields a domain error of type [E].
+     * @param ifSystemFailure the function to apply if [f] throws a [Throwable].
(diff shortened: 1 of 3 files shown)


joramnv:
See conversation in Slack for more context: https://kotlinlang.slack.com/archives/C5UPMM0A0/p1599853983245600

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -926,6 +926,60 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
...
+     * @param f the function that needs to be resolved.
+     * @param ifSuccess the function to apply if [f] yields a success of type [A].
+     * @param ifDomainError the function to apply if [f] yields a domain error of type [E].
+     * @param ifSystemFailure the function to apply if [f] throws a [Throwable].
+     * @param ifHandlingCaseThrows the function to apply if the [ifSuccess], [ifDomainError] or [ifSystemFailure] function throw a [Throwable].
+     * Throwing any [Throwable] in the [ifHandlingCaseThrows] function will render the [resolve] function nondeterministic.
+     * @param ifUnrecoverableState the function to apply if [resolve] is in an unrecoverable state.
+     * @return the result of applying the [resolve] function.
+     */
+    suspend fun <E, A, B> resolve(
```
raulraja, line 942: can we make all of these inline and add some docs to the Either and Fx docs regarding this new combinator?

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -926,6 +926,60 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
...
+      f: suspend () -> Either<E, A>,
+      ifSuccess: suspend (a: A) -> Either<Throwable, B>,
```
raulraja, line 944:
```suggestion
      success: suspend (a: A) -> Either<Throwable, B>,
```

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -926,6 +926,60 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
...
+      ifDomainError: suspend (e: E) -> Either<Throwable, B>,
```
raulraja, line 945:
```suggestion
      error: suspend (e: E) -> Either<Throwable, B>,
```

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -926,6 +926,60 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
...
+      ifSystemFailure: suspend (throwable: Throwable) -> Either<Throwable, B>,
```
raulraja, line 946:
```suggestion
      throwable: suspend (throwable: Throwable) -> Either<Throwable, B>,
```

raulraja (changes requested):

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
joramnv: Changed the naming suggestions using the IDE instead of committing the suggestion from GitHub. That way I could be sure that all references are updated as well.

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
joramnv: What exactly are you proposing to inline? 😯 
I'll look into adding some docs soon.

joramnv: Added some documentation. I can add some more documentation in a separate PR to the Arrow Fx docs. Perhaps (also) to the part about arrow-fx-coroutines-kotlinx-coroutines with a specific example of how to use the `resolve` function in Ktor routes.

raulraja: > What exactly are you proposing to inline? 😯 
Hi @joramnv, this is looking great, I ment inlining the resolve function itself.

joramnv: Hi Raúl, thank you. Your proposal to inline the resolve function is for performance reasons, right? Inlining it would mean fewer objects being instantiated and thus a smaller memory footprint and fewer objects that the GC has to deal with. Or are there other reasons to do this as well? (Just so that I understand why to inline. 😄)

I did change the function to be inlined, see last commit. But I had to mark some function parameters as noinline, because else the function could not use default values. Is this a desirable approach? Or should I investigate how to model the same behaviour without using default values (probably using function overloading).

By the way, I also did change the `catch` function to be inlined. I figured that the same applies to this function.

joramnv: I thought some more about it and I think this is a good compromise:
```
    suspend inline fun <E, A, B> resolve(
      f: suspend () -> Either<E, A>,
      success: suspend (a: A) -> Either<Throwable, B>,
      error: suspend (e: E) -> Either<Throwable, B>,
      throwable: suspend (throwable: Throwable) -> Either<Throwable, B>,
      unrecoverableState: suspend (throwable: Throwable) -> Either<Throwable, Unit>
    ): B =
      catch { f() }
        .fold({ t: Throwable -> throwable(t) }, { it.fold({ e: E -> handleItSafely { error(e) } }, { a: A -> handleItSafely { success(a) } }) })
        .fold({ t: Throwable -> throwable(t) }, { b: B -> b.right() })
        .fold({ t: Throwable -> unrecoverableState(t); throw t }, { b: B -> b })
```
The downsides I see are: less versatility (can not supply a function to specifically recover from success, error and throwable failures (but most often it would be the same function as the throwable function anyway)) and there now is a need to always supply an `unrecoverableState` function (which you probably want to do in most production code anyway and if you really want to do nothing at that point, you only need to add `Unit.right()` to the call sight). 

The upside is that the resolve function becomes fully inlined (no more noinline) and will profit from all performance benefits that inline functions bring. Also, if later on we decide that it is desired to have the possibility of a `handlingCaseThrows` like recovering functionality, there would still be a possibility to add an overloaded resolve function with such a function parameter.

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -1124,3 +1262,7 @@ inline fun <A, B> EitherOf<A, B>.handleErrorWith(f: (A) -> EitherOf<A, B>): Eith
       is Right -> it
     }
   }
+
+@PublishedApi
+internal suspend inline fun <A> handleItSafely(f: suspend () -> Either<Throwable, A>): Either<Throwable, A> =
```
raulraja, line 1267: Can this be another overload of Either.catch and become public?

raulraja (approved):
I think once CI is good this is ready, thanks so much @joramnv !

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
joramnv: With an overload like:
```
    @JvmName("catch1")
    suspend inline fun <R> catch(f: suspend () -> Either<Throwable, R>): Either<Throwable, R> =
      catch<Either<Throwable, R>>(f).fold({ it.left() }, { it })
```
The compiler gets confused: _Type mismatch: inferred type is Int but Either<Throwable, Int> was expected_ (This compile error message is for the "catch should return Right(result) when f does not throw" test case.).

Therefore, I choose to not overload but go with a different function name; `catchAndFlatten`. But the function is now more similar to the `catch` function and is public.

raulraja: Sounds good, thanks!

rachelcarmena:
Hi @joramnv :wave: Thank you so much for your contribution!! :raised_hands: 

It seems the change impacts on Arrow Fx. I think it won't be easy to find the reason because I didn't see real changes but additions here.

The failed CI checks test all the Arrow libraries with this pull request. In order to reproduce the error, please, run this command:

```
./gradlew buildWithLocalDeps
```

Instead of using Arrow dependencies from an external repository, that command will install Arrow dependencies in your local repository. You'll see something like:

```
arrow-core-lib [publishToMavenLocal]
...
arrow-ui-lib [publishToMavenLocal]
...
arrow-fx-lib [publishToMavenLocal]
...
(error)
```

I hope it's useful to detect the reason of the error. Please, let us know if we can help you :raised_hands:

joramnv:
> Hi @joramnv 👋 Thank you so much for your contribution!! 🙌
> 
> It seems the change impacts on Arrow Fx. I think it won't be easy to find the reason because I didn't see real changes but additions here.
> 
> The failed CI checks test all the Arrow libraries with this pull request. In order to reproduce the error, please, run this command:
> 
> ```
> ./gradlew buildWithLocalDeps
> ```
> 
> Instead of using Arrow dependencies from an external repository, that command will install Arrow dependencies in your local repository. You'll see something like:
> 
> ```
> arrow-core-lib [publishToMavenLocal]
> ...
> arrow-ui-lib [publishToMavenLocal]
> ...
> arrow-fx-lib [publishToMavenLocal]
> ...
> (error)
> ```
> 
> I hope it's useful to detect the reason of the error. Please, let us know if we can help you 🙌

Thanks for the pointer. Sadly, the `buildWithLocalDeps` task results in:

```
> Task :checkoutOrchestrator FAILED
Clone arrow repository in /Users/joram/projects/arrow ...
Cloning into 'arrow'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

FAILURE: Build failed with an exception.
```

joramnv:
`when (val res2 = Either.catch { scope.open(res.isInterruptible) }) {` (line 238 in `arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/Compiler.kt`) does not seem to be happy with `Either.catch` being an inline function. 

I see two possibilities; revert `Either.catch` back to not being inlined in this PR or change the implementation in the `arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/stream/Compiler.kt` (to use a separate not inlined local version of `Either.catch` or to make it work with the inline function).

raulraja:
Hi @joramnv , I think inlining the body in the stream compiler is the best option since Either.catch is heavily used and the Compiler impl is private. If we find a small repro for this outside Arrow we should consider reporting to the Kotlin tracker if isolated. Thanks!

rachelcarmena:
Thanks for the feedback @joramnv !! I'll add support for https as well :+1:

joramnv:
> Hi @joramnv , I think inlining the body in the stream compiler is the best option since Either.catch is heavily used and the Compiler impl is private. If we find a small repro for this outside Arrow we should consider reporting to the Kotlin tracker if isolated. Thanks!

Created a new PR for this: https://github.com/arrow-kt/arrow-fx/pull/287

nomisRev (approved):
Thanks for the great contribution @joramnv! 🎉

Issue #246, opened 2020-10-09
Refactor Arrow making use of Arrow meta features [Draft ticket]


danimontoya:
Since Arrow Meta is becoming stable we can start the integration into Arrow Core so we can make a nice refactor of current encoding by making use of all the Arrow meta features.

These are the first steps to get to the goal:

- [x] Enable the Arrow meta compiler plugin, so we can add it to the IDE. [Here is the setup guide](https://meta.arrow-kt.io/setup.html)
- [ ] Migrate the prelude module from meta into Arrow core, so we can make use of type proofs, Tuples, Unions, Refined types and so on.
- [ ] Refactor typeclasses. Make use of Coercion/Extension/Given type proofs. Start refactor with Semigroup.
- [ ] Deprecate current definition for Tuples and make use of the encoding done in Meta.
- [ ] Benchmark Union types vs Sealed classes
- [ ] More to be added here

PR #247, opened 2020-10-09
Arrow Meta integration within Arrow Core

danimontoya:
Addresses part of #246 .
Migration of prelude module from meta into Arrow core (migrate everything but `Kinds`)


arrow-core-data/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -544,11 +544,14 @@ sealed class Option<out A> : OptionOf<A> {
...
-    "Some(${SA.run { it.show() }})"
-  })
+  fun show(SA: Show<A>): String = fold(
+    {
+      "None"
+    },
+    {
+      "Some(${SA.run { it.show() }})"
+    }
+  )
```
rachelcarmena, line 554: Hi @danimontoya :raised_hands: I see several format changes. What if we move these format changes to a single PR (with just format changes) to detect what changes come from the integration in an easier way?

rachelcarmena: If they come from the new version of KTLint so we could create a PR for that upgrade before.

danimontoya: yeah, makes totally sense! I'll do that, thanks for the suggestion! 😄

arrow-annotations/build.gradle
```diff
@@ -9,3 +9,17 @@ apply from: "$DOC_CREATION"
 dependencies {
     api "io.kindedj:kindedj:$KINDEDJ_VERSION"
 }
+
+compileKotlin {
+    kotlinOptions {
+        useIR = true
+        freeCompilerArgs += ["-Xallow-jvm-ir-dependencies"]
```
rachelcarmena, line 16: @danimontoya :raised_hands: according to the errors, it seems this arg should be provided for all the modules. To solve it, you can create a PR in `arrow` repository with a branch called `new-arrow-meta-integration` (the same name) and include it here: https://github.com/arrow-kt/arrow/blob/master/gradle/subproject.gradle You'll see `kotlinOptions` there. Then re-run these checks.

raulraja:
On hold after 1.0

PR #251, opened 2020-10-15
Provide nullable continuation based on DelimitedScope

kioba:
Providing nullable continuation implementation after a [conversation](https://kotlinlang.slack.com/archives/C5UPMM0A0/p1602526947183200?thread_ts=1602526442.183000&cid=C5UPMM0A0) in the Kotlin Slack channel.

The `nullable` continuation will allow short circuit `null` values inside the block.

## Questions 

This is the first iteration with multiple questions regarding the implementation:

- `DelimContScope` used `null` to represent the end of the computation block but we need to keep the `null` value open for the results. The atomic values have been nested within the `Shift` sealed class to allow nullable values and also be able to represent the end of the block. is this effective enough or should we approach it from a different angle introducing an extra atomic value?
- To be able to use `DelimitedScope`, `arrow-core-data` needs to depend on `arrow-continuations`. Is this the right idea or should `nullable` continuation be part of the `arrow-continuations` module?
- A new `BindSyntax` has been introduced to allow the existing syntax to be used.example: 
```
nullable {
val a = "a".bind()
a.length.bind()
}
```
Should we stick to the `bind()` syntax or just simply use `DelimitedScope` and allow the user to use `shift` and `reset`?


raulraja:
Arrow core should depend on arrow-continuations 👍

raulraja:
Bind syntax should change to suspend operator fun invoke. We discussed simplifying the act of binding for suspended application of any data type to deprecate bind and ! and use just suspended invoke

arrow-core-data/src/main/kotlin/arrow/core/computations/nullable.kt
```diff
@@ -0,0 +1,27 @@
...
+    scope.shift { cont ->
+      this@bind?.let { cont(it) }
+    }
+}
+
+interface BindSyntax {
+  suspend fun <A> A?.bind(): A
+
+  suspend operator fun <A> A?.not(): A = bind()
+}
```
raulraja, line 26:
```suggestion
//this will become interface fun when they support suspend in the next release
interface BindSyntax { 
  suspend operator fun <A> A?.invoke(): A 
}
```

arrow-continuations/src/main/kotlin/arrow/continuations/generic/DelimContScope.kt
```diff
@@ -25,12 +25,12 @@ class DelimContScope<R>(val f: suspend DelimitedScope<R>.() -> R) : DelimitedSco
...
    * Variable used for polling the result after suspension happened.
    */
-  private val resultVar = atomic<R?>(null)
+  private val resultVar = atomic<Any?>(EMPTY_VALUE)
 
   /**
    * Variable for the next shift block to (partially) run, if it is empty that usually means we are done.
    */
-  private val nextShift = atomic<(suspend () -> R)?>(null)
+  private val nextShift = atomic<Any?>(EMPTY_VALUE)
```
1Jajen1, line 33: I don't actually think this is necessary here. Only the result var actually ever has values of type `R` and thus can lead to nested nulls, this one cannot.

arrow-core-data/build.gradle
```diff
@@ -11,6 +11,7 @@ apply from: "$DOC_CREATION"
 
 dependencies {
     api project(":arrow-annotations")
+    api project(":arrow-continuations")
```
1Jajen1, line 14: I don't think api visibility is a good idea for this as this isn't meant to be used by anyone using core directly.

arrow-core-data/src/main/kotlin/arrow/core/computations/suspended/NullableBindSyntax.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import arrow.continuations.generic.DelimitedScope
+
+data class NullableBindSyntax<R>(
+  private val scope: DelimitedScope<R?>,
+) : BindSyntax {
+  override suspend fun <A> A?.invoke(): A =
+    scope.shift { cont ->
+      this@invoke?.let { cont(it) }
+    }
```
1Jajen1, line 11: Can this be implemented to only call `shift` when `this` is `null` instead of calling it every time?

arrow-core-data/build.gradle
kioba: yeah, that is a good point! Will use `implementation` here 👍

arrow-core-data/src/main/kotlin/arrow/core/computations/suspended/NullableBindSyntax.kt
kioba: 🤔 `shift` promises to convert `A?` to `A` (nullable into a non-nullable). If my understanding is correct, scope.shift  is doing this with the continuation and if the value is null the continuation is not called.

kioba: if the `invoke()` is executed on a `null` value there will be no more invoke function called. am I assuming this correctly?

just the illustrate with an example:
```
nullable {
val a = null.invoke() // last swift call, short circuit to the end

val b = a.invoke() // wont be called because of short circuit 
b.invoke() // wont be called because of short circuit 
} == null
```

arrow-core-data/src/main/kotlin/arrow/core/computations/nullable.kt
kioba: Extended the comment with a //TODO just to allow discovery on the next release.

arrow-core-data/src/main/kotlin/arrow/core/computations/suspended/NullableBindSyntax.kt
1Jajen1: `shift` returns whatever is given to it's argument:
```kotlin
println(shift { it(3) }) // => 3
println(shift { it(9); it(5) }) // => 9\n5 (assuming multi shot is used)
```
So the return value of `shift` is what we pass to the continuation. But what if we never pass anything to it? Then it never returns anything: (The type of this is `Nothing` and infinite suspension, throwing and infinite recursion are the only ways to create a "value" of such a type. And since `Nothing` can be used in place of any type it does indeed also return `A`)
```kotlin
println(shift { _ -> null }) // => nothing here because we never resume
```
The function passed to `shift` has to return the final argument type of the delimited continuation `R` but how it does so doesn't actually matter. It is free to completely ignore the continuation passed to it.
```kotlin
val a: Int? = null
reset<Int?> {
  val b: Int = a ?: shift { null } // null is perfectly valid for `Int?` and `shift { null }` is `Nothing`
  b
} // => null
val a: Int? = null
reset<Int?> {
  val b: Int = a ?: shift { it(8) }
  b
} // => 8
```

There are a few things to note:
`shift { it(x) } == x` if, and only if, invoking the continuation is all you do (no other effects)
`shift { x }` has type `Nothing` since it short circuits
`shift { it(x); it(y) }` will first run the continuation with `x` and then with `y` (except we don't really support multishot so this will throw unless you use the multi shot continuation scope which, quite terribly, emulates this behavior)

1Jajen1: So the assumption that `shift` converts `A?` to `A` is wrong. It simply returns what is passed to the continuation. This is further evident by its type: `fun <A> shift(f: ((A) -> R) -> R): A`.

arrow-continuations/src/main/kotlin/arrow/continuations/generic/DelimContScope.kt
kioba: I fixed this in the meantime, you were right, I confused the nextShift with the result atomic. Thanks for pointing that out! 👍

arrow-core-data/src/main/kotlin/arrow/core/computations/suspended/NullableBindSyntax.kt
kioba: ok,it started to click. Do you think we could include this explanation in the documentation? 

now I see the reason behind what you mentioned. 
> only call shift when this is null

Will apply the changes quickly

raulraja: Jannis, excellent disection of shift, agreed  we need something like that in the docs.

arrow-core-data/src/main/kotlin/arrow/core/computations/suspended/NullableBindSyntax.kt
```diff
@@ -0,0 +1,10 @@
+package arrow.core.computations.suspended
+
+import arrow.continuations.generic.DelimitedScope
+
```
PhBastiani, line 4: Binding syntax for a [nullable] within a continuation context.

arrow-core-data/src/main/kotlin/arrow/core/computations/suspended/BindSyntax.kt
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Running A? in the context of [nullable]
+ *
+ * ```
+ * nullable {
+ *   val one = 1.invoke() // using invoke
+ *   val bigger = (one.takeIf{ it > 1 }).invoke() // using invoke on expression
+ *   bigger
+ * }
+ * ```
```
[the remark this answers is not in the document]
kioba: This version of BindSyntax only applies to the nullable implementation usually looking to use the Kinded version. also `NullableBindingSyntax` will be removed once fun interfaces are allowed to have suspendable functions.

arrow-core-data/src/main/kotlin/arrow/core/computations/suspended/NullableBindSyntax.kt
kioba: moved `NullableBindSyntax.kt` into `nullable` as a private class

nomisRev (approved):
Awesome work @kioba 😍

kioba:
> Awesome work @kioba 😍

Thank you, That means a lot for me! :)

1Jajen1 (approved):
Sorry for the late response :)

This looks good to me :+1:

kioba:
> Sorry for the late response :)
> 
> This looks good to me 👍

No worries, I really appreciate your help and guidance <3

nomisRev:
> Thank you, That means a lot for me! :)

❤️

PR #254, opened 2020-11-04
Make Either's handleErrorWith able to change error type

aballano:
In the function doc it says: "This is like `flatMap` for the exception.", but in reality is not because we cannot change the error type to something else, like we can do in flatMap for the right side.

Also fixes `contains` docs and simplifies `contains` and `exists` implementations.

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
@@ -996,8 +996,7 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
       return when (ev) {
         is Left -> Left(ev.a)
         is Right -> {
-          val b: Either<A, B> = ev.b
-          when (b) {
+          when (val b = ev.b) {
             is Left -> tailRecM(b.a, f)
             is Right -> Right(b.b)
           }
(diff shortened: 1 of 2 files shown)


nomisRev (approved):
👏

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -1261,7 +1261,7 @@ inline fun <A> Any?.rightIfNull(default: () -> A): Either<A, Nothing?> = when (t
  * Applies the given function `f` if this is a [Left], otherwise returns this if this is a [Right].
  * This is like `flatMap` for the exception.
  */
-inline fun <A, B> EitherOf<A, B>.handleErrorWith(f: (A) -> EitherOf<A, B>): Either<A, B> =
+inline fun <A, B, C> EitherOf<A, B>.handleErrorWith(f: (A) -> EitherOf<C, B>): Either<C, B> =
   fix().let {
     when (it) {
```
pakoito, line 1266: `when(val either = fix())`

arrow-core-data/src/test/kotlin/arrow/core/EitherTest.kt
```diff
@@ -229,6 +229,12 @@ class EitherTest : UnitSpec() {
       }
     }
 
+    "handleErrorWith should be able to change error type" {
+      forAll { a: Int, b: String ->
+        Left(a).handleErrorWith { Left(b) } == Left(b)
```
pakoito, line 234: One test for right with explicit types, just in case.

pakoito (approved):

PR #255, opened 2020-11-10
Add stm to quickstart

1Jajen1:
According to https://github.com/arrow-kt/arrow/blob/master/docs/libraries/how-to-add-a-new-module.md


rachelcarmena (approved):
:tada:

PR #258, opened 2020-11-29
SV. Correct suspension requires ControlThrowable

raulraja:


arrow-continuations/src/main/kotlin/arrow/continuations/generic/SuspendingComputation.kt
```diff
@@ -0,0 +1,112 @@
...
+import kotlinx.atomicfu.loop
+import kotlin.coroutines.Continuation
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlin.coroutines.intrinsics.COROUTINE_SUSPENDED
+import kotlin.coroutines.intrinsics.startCoroutineUninterceptedOrReturn
+import kotlin.coroutines.intrinsics.suspendCoroutineUninterceptedOrReturn
+import kotlin.coroutines.resumeWithException
+
+sealed class Either<out E, out A>
```
raulraja, line 13: I guess we can remove this sample code for Either?

arrow-continuations/src/main/kotlin/arrow/continuations/generic/SuspendingComputation.kt
```diff
@@ -0,0 +1,112 @@
...
+}
+
+suspend fun <E, A> either(f: suspend EitherBind<E>.() -> A): Either<E, A> =
+  suspendCoroutineUninterceptedOrReturn { cont ->
+    SuspendMonadContinuation(cont).startCoroutineUninterceptedOrReturn {
+      f.invoke(object : EitherBind<E> {
+        override suspend fun <B> Either<E, B>.invoke(): B =
+          when(val ea = this) {
+            is Right -> ea.a
+            is Left -> shift(ea)
```
raulraja, line 28: awesome!

arrow-core-data/src/main/kotlin/arrow/core/NonFatal.kt
```diff
@@ -41,7 +43,7 @@ package arrow.core
  */
 fun NonFatal(t: Throwable): Boolean =
   when (t) {
-    is VirtualMachineError, is ThreadDeath, is InterruptedException, is LinkageError -> false
+    is VirtualMachineError, is ThreadDeath, is InterruptedException, is LinkageError, is ControlThrowable -> false
```
raulraja, line 46: 👍

arrow-continuations/src/main/kotlin/arrow/continuations/generic/SuspendingComputation.kt
```diff
@@ -0,0 +1,112 @@
...
+
+
+interface SuspendingComputation<R> {
+  suspend fun <A> shift(a: R): A
+}
+
+internal const val UNDECIDED = 0
+internal const val SUSPENDED = 1
+
+class ShortCircuit(val raiseValue: Any?) : ControlThrowable()
```
raulraja, line 42: If we move this to an inner class inside `SuspendingComputation` we can type this such as :
```
interface SuspendingComputation<R> {
  inner class ShortCircuit(val value: R) : ControlThrowable
  suspend fun <A> shift(a: R): A
}
```
This way we can ensure they can only short-circuit to valid values for that context

raulraja (commented):
@nomisRev Can't approve this since I raised the PR from your branch but if it looks good we can merge into the branch that is pending for master

arrow-continuations/src/main/kotlin/arrow/continuations/generic/SuspendingComputation.kt
nomisRev: Sounds good!

arrow-continuations/src/main/kotlin/arrow/continuations/generic/SuspendingComputation.kt
nomisRev: Yes, can be removed just serverd as a placeholder type.

arrow-continuations/src/main/kotlin/arrow/continuations/generic/SuspendingComputation.kt
```diff
@@ -0,0 +1,112 @@
...
+  @PublishedApi // return the result
+  internal fun getResult(): Any? =
+    _decision.loop { decision ->
+      when (decision) {
+        UNDECIDED -> if (this._decision.compareAndSet(UNDECIDED, SUSPENDED)) return COROUTINE_SUSPENDED
+        else -> return decision
+      }
+    }
+
+  override suspend fun <A> shift(a: R): A = throw ShortCircuit(a)
```
raulraja, line 100: if shift is going to be implemented like this it can be moved to the interface in which case the method does not need suspension maybe?

nomisRev: Yes, and it doesn't need suspension.

nomisRev: We need to check/add support for nested scopes tho.

nomisRev: We can do that here or in the other PR.

arrow-continuations/src/main/kotlin/arrow/continuations/generic/SuspendingComputation.kt
nomisRev: Impossible: `Inner class of generic class extending 'Throwable' is prohibited`

arrow-continuations/src/main/kotlin/arrow/continuations/generic/SuspendingComputation.kt
nomisRev: Added in this PR with tests.

nomisRev: @raulraja I'm not sure which interface you mean? It's already in an interface 🤔

1Jajen1: > Yes, and it doesn't need suspension.

There is only one use case that benefits from suspension that I can think of and that is a delayed resumption of the continuation. Some datatypes like `Eval` need that when binding, but all eager types where its either immediate resumption or short circuit don't need suspension at all and are likely much better off without it.

raulraja:
replaced by #259

PR #259, opened 2020-11-30
Restricted & Suspended delimited scopes

raulraja:
- `Reset.restricted` and `Reset.suspended`.
- Introduction of `Effect` polymorphic over both.
- Rewire `either` and `nullable` to be based on `Effect`
- Suspended scope shifts with `ControlThrowable`
- Restricted scope implements traditional `reset/shift` with continuation callback instead of `ControlThrowable`.

A joint effort with @nomisRev @1Jajen1 and @rachelcarmena


arrow-core-data/src/main/kotlin/arrow/core/computations/either.kt
```diff
@@ -1,38 +1,22 @@
...
-          override suspend fun <A> Kind<EitherPartialOf<E>, A>.invoke(): A =
-            when (val v = fix()) {
-              is Either.Right -> v.b
-              is Either.Left -> shift { v }
-            }
-        })
-      )
-    }
+object either {
+  inline fun <E, A> eager(crossinline c: suspend EitherEffect<E, *>.() -> A): Either<E, A> =
```
nomisRev, line 17: Idem as `nullable`.

arrow-core-data/src/main/kotlin/arrow/core/computations/nullable.kt
```diff
@@ -1,18 +1,17 @@
...
+object nullable
 
-  private class NullableBindSyntax<R>(
-    scope: DelimitedScope<R?>
-  ) : BindSyntax, DelimitedScope<R?> by scope {
-    override suspend fun <A> A?.invoke(): A =
-      this ?: shift { null }
-  }
-}
+fun <A> nullable.eager(func: suspend NullableEffect<A>.() -> A?): A? =
```
nomisRev, line 13: Allows for foreign extension. We need to define this seperately, and then it's safe.

```
@RestrictsSuspension
fun interface EagerNullableEffect<A> : Effect<A?> {
  suspend operator fun <B> B?.invoke(): B = this ?: control().shift(null)
}
```

```suggestion
fun <A> nullable.eager(func: suspend EagerNullableEffect<A>.() -> A?): A? =
```

arrow-core-data/src/main/kotlin/arrow/core/computations/suspended/BindSyntax.kt
```diff
@@ -12,6 +14,11 @@ package arrow.core.computations.suspended
...
  */
 // TODO: this will become interface fun when they support suspend in the next release
+@RestrictsSuspension
+interface EagerBindSyntax {
+  suspend operator fun <A> A?.invoke(): A
+}
+
 interface BindSyntax {
   suspend operator fun <A> A?.invoke(): A
 }
```
nomisRev, line 24: These can be removed in favor of the ones directly in `nullable.kt`

nomisRev (commented):
Looks good. The `eager` `either` and `nullable` computation blocks still need a `@RestrictSuspension` receiver syntax interface.

arrow-continuations/build.gradle
```diff
@@ -12,4 +12,9 @@ dependencies {
     testRuntimeOnly "org.junit.vintage:junit-vintage-engine:$JUNIT_VINTAGE_VERSION"
     testCompileOnly "io.kotlintest:kotlintest-runner-junit5:$KOTLIN_TEST_VERSION", excludeArrow
     testImplementation project(":arrow-core-test")
+    testImplementation "io.arrow-kt:arrow-fx-coroutines:$VERSION_NAME"
+}
+
+ktlint {
+    version = "0.39.0"
```
rachelcarmena, line 19:
```suggestion
```

It's already defined in the main configuration.

rachelcarmena (approved):
Awesome, thanks @raulraja :raised_hands:

nomisRev (approved):
We should deprecate `validated` since it makes no sense in a short-circuit scenario. We could alternatively offer `Validated` syntax in `either` instead.

We should clarify this clearly in the documentation, but that can both be taken in a different PR IMO.

raulraja:
@nomisRev I'll quickly add this here too since it's not merged yet and we take care of it

nomisRev (commented):
🚀

PR #266, opened 2020-12-12
Adds deprecated bind operators to Effect impls

raulraja:
Adds the deprecated operators we had in BindSyntax in order to offer a deprecation cycle in 0.12 and to be removed in 0.13

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/core/computations/const.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/computations/const.kt
@@ -1,0 +1,39 @@
+package arrow.core.computations
+
+import arrow.continuations.Effect
+import arrow.core.Const
+import arrow.core.const
+import kotlin.coroutines.RestrictsSuspension
+
+@Deprecated(
+  "Const binding doe not require suspension and this computation block will be removed."
+)
+fun interface ConstEffect<A, T> : Effect<Const<A, T>> {
+
+  @Deprecated("The monadic operator for the Arrow 1.x series will become invoke in 0.13", ReplaceWith("()"))
+  suspend fun <B> Const<A, B>.bind(): B = this()
+
+  @Deprecated("The monadic operator for the Arrow 1.x series will become invoke in 0.13", ReplaceWith("()"))
+  suspend operator fun <B> Const<A, B>.not(): B = this()
+
+  suspend operator fun <B> Const<A, B>.invoke(): B =
+    value() as B
+}
+
+@Deprecated(
+  "Const binding doe not require suspension and this computation block will be removed."
+)
+@RestrictsSuspension
+fun interface RestrictedConstEffect<E, A> : ConstEffect<E, A>
+
+@Suppress("ClassName")
+@Deprecated(
+  "Const binding doe not require suspension and this computation block will be removed."
+)
+object const {
+  inline fun <A, T> eager(crossinline c: suspend RestrictedConstEffect<A, *>.() -> A): Const<A, T> =
+    Effect.restricted(eff = { RestrictedConstEffect { it } }, f = c, just = { it.const() })
+
+  suspend inline operator fun <A, T> invoke(crossinline c: suspend ConstEffect<A, *>.() -> A): Const<A, T> =
+    Effect.suspended(eff = { ConstEffect { it } }, f = c, just = { it.const() })
+}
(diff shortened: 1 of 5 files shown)


nomisRev (commented):
Missing for the deprected `validated` block. I saw we also removed `const` we should also bring that back for deprecation.

raulraja:
@nomisRev, added those back

nomisRev (approved):
Thanks @raulraja 🙌 😘

arrow-core-data/src/main/kotlin/arrow/core/computations/const.kt
```diff
@@ -0,0 +1,39 @@
+package arrow.core.computations
+
+import arrow.continuations.Effect
+import arrow.core.Const
+import arrow.core.const
+import kotlin.coroutines.RestrictsSuspension
+
+@Deprecated(
+  "Const binding doe not require suspension and this computation block will be removed."
```
i-walker, line 9:
```suggestion
  "Const binding does not require suspension and this computation block will be removed."
```

arrow-core-data/src/main/kotlin/arrow/core/computations/const.kt
```diff
@@ -0,0 +1,39 @@
...
+
+  @Deprecated("The monadic operator for the Arrow 1.x series will become invoke in 0.13", ReplaceWith("()"))
+  suspend operator fun <B> Const<A, B>.not(): B = this()
+
+  suspend operator fun <B> Const<A, B>.invoke(): B =
+    value() as B
+}
+
+@Deprecated(
+  "Const binding doe not require suspension and this computation block will be removed."
```
i-walker, line 24:
```suggestion
  "Const binding does not require suspension and this computation block will be removed."
```

arrow-core-data/src/main/kotlin/arrow/core/computations/const.kt
```diff
@@ -0,0 +1,39 @@
...
+)
+@RestrictsSuspension
+fun interface RestrictedConstEffect<E, A> : ConstEffect<E, A>
+
+@Suppress("ClassName")
+@Deprecated(
+  "Const binding doe not require suspension and this computation block will be removed."
```
i-walker, line 31:
```suggestion
  "Const binding does not require suspension and this computation block will be removed."
```

i-walker (approved):
👍 Thanks @raulraja

arrow-core-data/src/main/kotlin/arrow/core/computations/const.kt
```diff
@@ -0,0 +1,39 @@
...
+)
+fun interface ConstEffect<A, T> : Effect<Const<A, T>> {
+
+  @Deprecated("The monadic operator for the Arrow 1.x series will become invoke in 0.13", ReplaceWith("()"))
+  suspend fun <B> Const<A, B>.bind(): B = this()
+
+  @Deprecated("The monadic operator for the Arrow 1.x series will become invoke in 0.13", ReplaceWith("()"))
+  suspend operator fun <B> Const<A, B>.not(): B = this()
+
+  suspend operator fun <B> Const<A, B>.invoke(): B =
```
JorgeCastilloPrz, line 19: Not to open a discussion again, but something I'm not a big fan of is how invoke looks when working with the usual computations that return a monad: 

```kotlin
either {
  val speaker = service.loadSpeaker()() // or service.loadSpeaker().invoke()
  val talks = service.loadTalks(speaker.talkIds)()
}
```

This is likely how programs are gonna look overall unless I'm missing something. You end up having the double parenthesis everywhere. Also, just for relevance: On all courses we show bind as `!` or explicit `.bind()`, sometimes we also mentioned the style using destructuring. Every time they have picked the explicit one: ".bind()". The not operator is the one less embraced overall.

rachelcarmena: Hi Jorge, thanks!! :raised_hands: Discussions have been enabled in `arrow` repository in order not to miss these things: https://github.com/arrow-kt/arrow/discussions

raulraja: Invoke was chosen because it offers us both options whereas bind is not an operator so the code is all calls to .bind() With `invoke` the user can choose an explicit name or operator. I prefer user invoke but if we want to change it we are still in time before 0.12.x

nomisRev: https://kotlinlang.slack.com/archives/C5UPMM0A0/p1608320480156200

PR #267, opened 2020-12-14
Deprecate @higherkind & @extension for Validated and offer concrete implementations

nomisRev:
This PR deprecates @higherkind and @extension for Validated and offers concrete signatures for all functions exposed as `@extensions`.

- [x] Temporarly move generated code to arrow-core, and disable `@extensions` generation
- [x] Offer concrete implementations for `@extensions` generated code
- [x] Deprecate all `@extensions` generated code with proper `ReplaceWith` methods.

This PR also removes `Functor#unit` which has been deprecated in favor of `void` since `0.10.x`

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
@@ -869,7 +869,6 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
    * Left(12).map { "flower" }  // Result: Left(12)
    * ```
    */
-  @Suppress("UNCHECKED_CAST")
   inline fun <C> map(f: (B) -> C): Either<A, C> =
     flatMap { Right(f(it)) }
 
@@ -996,6 +995,12 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
 
     fun <R> right(right: R): Either<Nothing, R> = Right(right)
 
+    val leftUnit: Either<Unit, Nothing> = left(Unit)
+
+    val unit: Either<Nothing, Unit> = right(Unit)
+
+    fun <L> unit(): Either<L, Unit> = unit
+
     fun <A> fromNullable(a: A?): Either<Unit, A> = a?.right() ?: Unit.left()
 
     tailrec fun <L, A, B> tailRecM(a: A, f: (A) -> Kind<EitherPartialOf<L>, Either<A, B>>): Either<L, B> {
@@ -1077,6 +1082,12 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
         .fold({ t: Throwable -> throwable(t) }, { b: B -> b.right() })
         .fold({ t: Throwable -> unrecoverableState(t); throw t }, { b: B -> b })
   }
+
+  fun <C> mapConst(b: C): Either<A, C> =
+    map { b }
+
+  fun void(): Either<A, Unit> =
+    mapConst(Unit)
 }
 
 fun <L> Left(left: L): Either<L, Nothing> = Left(left)
--- a/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
@@ -1083,8 +1083,8 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
         .fold({ t: Throwable -> unrecoverableState(t); throw t }, { b: B -> b })
   }
 
-  fun <C> mapConst(b: C): Either<A, C> =
-    map { b }
+  fun <C> mapConst(c: C): Either<A, C> =
+    map { c }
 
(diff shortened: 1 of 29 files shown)


arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -1077,6 +1082,12 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
         .fold({ t: Throwable -> throwable(t) }, { b: B -> b.right() })
         .fold({ t: Throwable -> unrecoverableState(t); throw t }, { b: B -> b })
   }
+
+  fun <C> mapConst(b: C): Either<A, C> =
+    map { b }
```
pablisco, line 1087: Should we rename the parameter as `c`?

arrow-core/src/main/kotlin/arrow/core/extensions/validated/applicative/ValidatedApplicative.kt
```diff
@@ -0,0 +1,85 @@
...
+): Validated<E, A> = fix()._replicate(SE, arg1, arg2)
+
+@Suppress(
+  "UNCHECKED_CAST",
+  "NOTHING_TO_INLINE"
+)
+@Deprecated("Applicative typeclasses is deprecated. Use concrete methods on Validated")
+inline fun <E> Companion.applicative(SE: Semigroup<E>): ValidatedApplicative<E> = object :
+    arrow.core.extensions.ValidatedApplicative<E> { override fun SE():
+    arrow.typeclasses.Semigroup<E> = SE }
```
pablisco, line 85:
```suggestion
    ValidatedApplicative<E> { override fun SE(): Semigroup<E> = SE }
```

arrow-core/src/main/kotlin/arrow/core/extensions/validated/applicative/ValidatedApplicative.kt
```diff
@@ -0,0 +1,85 @@
...
+fun <E, A> Kind<Kind<ForValidated, E>, A>.replicate(SE: Semigroup<E>, arg1: Int): Validated<E, List<A>> =
+  fix()._replicate(SE, arg1)
+
+@JvmName("replicate")
+@Suppress(
+  "UNCHECKED_CAST",
+  "USELESS_CAST",
+  "EXTENSION_SHADOWED_BY_MEMBER",
+  "UNUSED_PARAMETER"
+)
```
pablisco, line 70: Is it worth moving this suppression to the file? (`@file:Supress(...)`)

arrow-core/src/main/kotlin/arrow/core/extensions/validated/applicativeError/ValidatedApplicativeError.kt
```diff
@@ -0,0 +1,181 @@
...
+@Suppress(
+  "UNCHECKED_CAST",
+  "NOTHING_TO_INLINE"
+)
+@Deprecated("ApplicativeError typeclasses is deprecated. Use concrete methods on Validated")
+inline fun <E> Companion.applicativeError(SE: Semigroup<E>): ValidatedApplicativeError<E> = object :
+  arrow.core.extensions.ValidatedApplicativeError<E> {
+  override fun SE():
+    arrow.typeclasses.Semigroup<E> = SE
+}
```
pablisco, line 181:
```suggestion
  ValidatedApplicativeError<E> { override fun SE(): Semigroup<E> = SE }
```

pablisco: I realised that this happens on all types, shall I create a separate branch with changes or we don't care that much as it's gonna go away anyway 😅

pablisco (commented):
Looking good :) I realised half way through that this may be somewhat generated, so feel free to ignore some of my comments if they aren't relevant

nomisRev:
@pablisco Yes, I don't think we really care too much about the generated code. It's all `@Deprecated` and we hope to remove it asap.

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
nomisRev: Yes, we should. Should probably also re-check the `Validated.kt` extension functions since some are using `<A, B> and some are using `<E, A>` and others <L, R>.

pablisco (approved):

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
nomisRev: Updated.

rachelcarmena (approved):
Thanks!! :raised_hands: Using `PublishedApi` :clap:

raulraja (approved):
Looks great, thanks @nomisRev

PR #268, opened 2020-12-14
Deprecate BindingSyntax and remove usages of deprecated operators

i-walker:
Deprecates BindingSyntax in favor of Invoke


i-walker:
I'm unsure if we want to delete BindingSyntax in this SNAPSHOT or in the next one. 
I am open to doing it now if required.

nomisRev:
Closing this PR since most of these docs are going to be removed or rewritten by the time we remove `component1` and `not`.

PR #269, opened 2020-12-15
Bugfix suspended computation blocks

nomisRev:
1. KotlinX Coroutines throws `CancellationException` with `cause` when propagating cancellation by exception from `async` to it's parent. Since we cause the `CancellationException` with `ShortCircuit` we need to look if `ShortCircuit` can be found within the causes to return the excepted shifted value. If the cancellation is immediate, then it doesn't wrap in `CancellationException`, that caused the test to be flaky. Now it passes consitently on a loop.

2. Suspended computation blocks were running on `EmptyCoroutineContext`, they need to run on the parents `ctx` to correctly propogate  the state from outside to  inside the computation block.


raulraja (approved):
Thanks @nomisRev !

i-walker (approved):
Thanks, @nomisRev for the nice explanation!

PR #272, opened 2020-12-16
Deprecate @higherkind & @extension for Either

nomisRev:
- [x] Eq (EqK, EqK2)
- [x] Order
- [x] Hash
- [x] Show
- [x] Semigroup (SemigroupK)
- [x] Monoid
- [x] Functor    
- [x] Apply
- [x] Applicative
- [x] ApplicativeError
- [x] Monad
- [x] MonadError
- [x] Traverse
- [x] BiTraverse
- [x] Foldable
- [x] BiFoldable
- [x] BiFunctor
- [x] BiCrosswalk

Needs to be merged together with:
 - https://github.com/arrow-kt/arrow-incubator/pull/140
 - https://github.com/arrow-kt/arrow-fx/pull/354

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
@@ -1240,8 +1240,9 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
      *
      *  fun main(args: Array<String>) {
      *   //sampleStart
-     *   val f = Either.lift<String, String, String>({ s: CharSequence -> "$s World" })
-     *   val result = f("Hello".right())
+     *   val f = Either.lift<Int, CharSequence, String> { s: CharSequence -> "$s World" }
+     *   val either: Either<Int, CharSequence> = "Hello".right()
+     *   val result = f(either)
      *   //sampleEnd
      *   println(result)
      *  }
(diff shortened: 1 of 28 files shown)


arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -1041,24 +1162,52 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
     inline fun <L, R> conditionally(test: Boolean, ifFalse: () -> L, ifTrue: () -> R): Either<L, R> =
       if (test) right(ifTrue()) else left(ifFalse())
 
+    @Deprecated("Use the inline version. Hidden for binary compat", level = DeprecationLevel.HIDDEN)
     suspend inline fun <R> catch(f: suspend () -> R): Either<Throwable, R> =
+      catch { f() }
+
+    inline fun <R> catch(f: () -> R): Either<Throwable, R> =
```
nomisRev, line 1169: This breaks the API in some places where you pass a `suspend` lambda directly to the `catch` function, I think the Kotlin compiler has the same limitation for suspend method references..

See https://github.com/arrow-kt/arrow-fx/pull/354 as example of this breakage.

1Jajen1 (approved):
Looks good to me, thanks! :clap: 

So our approach for now is copy the (deprecated) kapt output, remove kapt, add meta?

i-walker (approved):
🙌 Thanks Simon

1Jajen1:
Looking at how https://github.com/arrow-kt/arrow-fx/pull/354 fails this seems to cause some other problems as well :/

nomisRev:
> So our approach for now is copy the (deprecated) kapt output, remove kapt, add meta?

Yes exactly. That's the most graceful way to get rid of the things we want to get rid of for 1.0. Almost all signatures can be preserved, except `traverse` which relies on `Applicative`. So that results in a couple of signatures instead, which need to be differentiated by function name atm due to a bug in the compiler where it cannot differentiate on the return type of the lambda.

1Jajen1:
> due to a bug in the compiler where it cannot differentiate on the return type of the lambda.

I think this is fixable in some cases with a custom `@jvmName` iirc.

nomisRev:
@1Jajen1 I originally thought so too, but we asked in #compiler on Kotlin Slack and doesn't seem to be the case. See https://youtrack.jetbrains.com/issue/KT-43872

PR #274, opened 2020-12-18
Inline new Validated functions

nomisRev:
Some new functions from Validated where not inlined, this PR inlines the missed functions


i-walker (approved):
👍 Thanks, Simon

1Jajen1 (approved):

PR #275, opened 2020-12-19
Deprecate IO (core docs)

raulraja:
In preparation of the IO deprecation this PR removes and adapts examples using `IO` from the core repo.


arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -340,41 +334,36 @@ import arrow.core.ValidatedNel
...
+ * suspend fun List<User>.processLogin(): Either<NotFound, List<Profile>> =
+ *   either {
+ *     parTraverse { userInfo(it)() }
+ *   }
  *
- * fun program(): IO<Unit> = IO.fx {
+ * suspend fun program(): Unit {
  *   val list = listOf(DummyUser("Micheal"), DummyUser("Juan"), DummyUser("T'Challa"))
- *     .processLogin()()
+ *     .processLogin()
```
i-walker, line 360: Doesn't this have to be invoked?

raulraja: Here we want to show the result of `processLogin` which is a `Either<NotFound, List<Profile>>`

arrow-docs/docs/patterns/glossary/README.md
```diff
@@ -403,6 +392,6 @@ Side-effects are too general to be unit tested for because they depend on the en
...
 
-For more complicated side-effects that can throw or jump threads, we need more advanced datatypes, called Effects, that wrap over impure operations. Some of these datatypes may already be familiar to you, like [`rx.Observable`]({{ '/integrations/rx2/' | relative_url }}), [`kotlinx.coroutines.Deferred`]({{ '/integrations/kotlinxcoroutines/' | relative_url }}), or Arrow's [`IO`]({{ '/effects/io/' | relative_url }}). These Effects compose, catch exceptions, control asynchrony, and, most importantly, can be run lazily. This gets rid of the issues with side-effects.
+For more complicated effects that can throw or jump threads, we need more advanced techniques. We model side-effects in kotlin with `suspend` functions and Kotlin Continuations. These continuations compose, catch exceptions, control asynchrony, and, most importantly, can be run lazily. This gets rid of the issues with side-effects.
```
i-walker, line 395: 🙌

i-walker (approved):
Thanks @raulraja, left only one question

arrow-docs/docs/datatypes/intro/README.md
```diff
@@ -1,170 +0,0 @@
----
```
pakoito, line 1: All gone? :(

raulraja: Not actually this one does not need to go if I can adapt it to Either but this tutorial will need to be rewritten based on some other technique that is not abstracting away MonadError since those type classes that are kinded are going away. I'll try to bring it back

raulraja: restored!

nomisRev (approved):
Awesome thanks for cleaning this up!

PR #277, opened 2020-12-21
Remove mtl dependency from Arrow Core Data

nomisRev:


raulraja (approved):
thanks Simon!

rachelcarmena (approved):
Thanks @nomisRev for this beautiful Christmas gift!! It will be super useful for the next steps!!

PR #276, opened 2020-12-19
Deprecate @higherkind & @extension for ListK

nomisRev:
- [x] Eq (EqK, EqK2)
- [x] Order
- [x] Hash
- [x] Show
- [x] Semigroup (SemigroupK)
- [x] Monoid (MonoidK)
- [x] Functor    
- [x] Apply
- [x] Applicative
- [x] Monad
- [x] Traverse
- [x] Foldable
- [x] Crosswalk
- [x] MonadLogic
- [x] MonadPlus
- [x] Alternative
- [x] Align
- [x] MonadFilter
- [x] MonadCombine
- [x] Monoidal
- [x] Semigroupal
- [x] Semialign
- [x] Unalign
- [x] Zip
- [x] Unzip

PR #278, opened 2020-12-21
Move nullable mapN to Nullable, and deprecate top-level ones

nomisRev:
Currently we cannot define `mapN` for `List` or `Map` since they don't have `companion object` in the Std, and they conflict with the `A?` nullable version of `mapN`.

To solve this we need to move the top-level `nullable` extension functions to `object Nullable`, this allows us to write extensions for `nullable` in a way they don't pollute the namespace for all other types.

See discussion: https://github.com/arrow-kt/arrow-core/pull/276#discussion_r546416592 for more details.

cc\\ original author @tapegram


i-walker (approved):
Thanks @nomisRev 😄

raulraja (approved):

PR #279, opened 2020-12-21
Automatically deprecate all `@extension` projected methods.

raulraja:
Still trying to determine if there is a way to actually find non implemented methods so they can be reported to be implemented. As is this already deprecates automatically all the generated extensions


arrow-meta/src/main/java/arrow/core/extensions/ExtensionProcessor.kt
```diff
@@ -223,6 +227,12 @@ class ExtensionProcessor : MetaProcessor<extension>(extension::class), PolyTempl
                 """"USELESS_CAST"""",
                 """"EXTENSION_SHADOWED_BY_MEMBER"""",
                 """"UNUSED_PARAMETER""""
+              ),
+              DeprecatedAnnotation(
+                "@extension kinded projected functions are deprecated",
+                "${func.name}(${allArgs.joinToString(", ") { it.name.trim() }})",
+                if (func.receiverType != null) listOf("arrow.core.${func.name}")
+                else listOf("${+companionOrFactory}.${func.name}")
```
raulraja, line 235: Here we can choose if we want the user code replacement by default to to look like "Option.just" or :
```
import arrow.core.Option.just

val x = just(1) // Option(1)
```

franciscodr (approved):

nomisRev (approved):

PR #280, opened 2020-12-22
Adding Iter interface

franciscodr:
This pull request adds a new interface called Iter that will allow the data types to implement the foldMap method for Traversal in Arrow Optics easily once the Kind type is removed.

It also provide an implementation of the interface for Either, Option and Validated


raulraja (approved):

raulraja:
Closed since after further discussion Traverse is gonna loose the foldMap related methods which will get moved to Each, constructed with both a fold and map

PR #283, opened 2020-12-23
Deprecate @higherkind and @extension for Ordering

franciscodr:
- [x] Eq
- [x] Hash
- [x] Monoid
- [x] Order
- [x] Semigroup
- [x] Show

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/core/Ordering.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Ordering.kt
@@ -116,7 +116,7 @@ private object OrderingHash : Hash<Ordering> {
 
 @Suppress("EXTENSION_SHADOWED_BY_MEMBER")
 private object OrderingMonoid : Monoid<Ordering> {
-  override fun empty(): Ordering = this.empty()
+  override fun empty(): Ordering = EQ
 
   override fun Ordering.combine(b: Ordering): Ordering =
     this.combine(b)
--- a/arrow-core-data/src/main/kotlin/arrow/core/Ordering.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Ordering.kt
@@ -1,5 +1,13 @@
 package arrow.core
 
+import arrow.typeclasses.Eq
+import arrow.typeclasses.Hash
+import arrow.typeclasses.Monoid
+import arrow.typeclasses.Order
+import arrow.typeclasses.Semigroup
+import arrow.typeclasses.Show
+import arrow.typeclasses.hashWithSalt
+
 sealed class Ordering {
   override fun equals(other: Any?): Boolean = this === other // ref equality is fine because objects should be singletons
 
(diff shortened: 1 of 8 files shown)


i-walker (approved):
Thanks @franciscodr 👍

raulraja:
I believe we should get rid of `Ordering` unless it brings something to the table and deprecate the main function in `Order` in favor of `operator fun compareTo(other: T): Int`. 
Is there a use case for Ordering that this operator does not already cover when implemented concretely on every data type? 
If there isn't I believe `Order` and `Ordering` should be deprecated in 0.12.0 and removed in 0.13.0.

1Jajen1:
I disagree with removing it in favor of `Int` based order apis. I find `fun compareTo(other: T): Int` very confusing and repeatedly have to lookup how exactly it works. It also does not compose at all whereas `Ordering` forms a lawful monoid.
We also provide very easy interop with `Int` based orderings in `Order` and easy conversions between `Ordering` and `Int`.

Although to be fair both `Order` and `Ordering` will likely see little use, but on the other hand it is also quite low maintenance, so I think we could keep it.

raulraja:
@1Jajen1 sounds good to keep Ordering but I propose we make one small change to the Order type class. 
`Order` should be a fun interface and the abstract method should be the one for the operator.
This is so we can construct order instances with no cost using the existing operators that regardless of Order or Ordering will appear concrete in the data type. This operators will appear concrete regardless so users can use not the the syntax `>`, `<=`, ... etc over Option, Either and others that can provide the operator.

arrow-core/src/main/kotlin/arrow/core/extensions/ordering/hash/OrderingHash.kt
```diff
@@ -0,0 +1,34 @@
...
+@Deprecated("@extension kinded projected functions are deprecated", ReplaceWith("hashWithSalt(arg1)"))
+fun Ordering.hashWithSalt(arg1: Int): Int = arrow.core.Ordering.hash().run {
+  this@hashWithSalt.hashWithSalt(arg1) as kotlin.Int
+}
+
+@Suppress(
+  "UNCHECKED_CAST",
+  "NOTHING_TO_INLINE"
+)
+@Deprecated("@extension kinded projected functions are deprecated", ReplaceWith("Ordering.hash()"))
```
nomisRev, line 33:
```suggestion
@Deprecated("@extension kinded projected functions are deprecated", ReplaceWith("Ordering.hash()", "arrow.core.hash"))
```

arrow-core/src/main/kotlin/arrow/core/extensions/ordering/monoid/OrderingMonoid.kt
```diff
@@ -0,0 +1,46 @@
...
+internal val monoid_singleton: OrderingMonoid = object : arrow.core.extensions.OrderingMonoid {}
+
+@JvmName("combineAll")
+@Suppress(
+  "UNCHECKED_CAST",
+  "USELESS_CAST",
+  "EXTENSION_SHADOWED_BY_MEMBER",
+  "UNUSED_PARAMETER"
+)
+@Deprecated("@extension kinded projected functions are deprecated", ReplaceWith("combineAll()", "arrow.core.combineAll"))
```
nomisRev, line 25: This is not defined?

arrow-core/src/main/kotlin/arrow/core/extensions/ordering/order/OrderingOrder.kt
```diff
@@ -0,0 +1,132 @@
...
+internal val order_singleton: OrderingOrder = object : arrow.core.extensions.OrderingOrder {}
+
+@JvmName("compareTo")
+@Suppress(
+  "UNCHECKED_CAST",
+  "USELESS_CAST",
+  "EXTENSION_SHADOWED_BY_MEMBER",
+  "UNUSED_PARAMETER"
+)
+@Deprecated("@extension kinded projected functions are deprecated", ReplaceWith("compareTo(arg1)", "arrow.core.compareTo"))
```
nomisRev, line 26: These can all be concrete and then don't require imports

arrow-core/src/main/kotlin/arrow/core/extensions/ordering/semigroup/OrderingSemigroup.kt
```diff
@@ -0,0 +1,46 @@
...
+    arrow.core.extensions.OrderingSemigroup {}
+
+@JvmName("plus")
+@Suppress(
+  "UNCHECKED_CAST",
+  "USELESS_CAST",
+  "EXTENSION_SHADOWED_BY_MEMBER",
+  "UNUSED_PARAMETER"
+)
+@Deprecated("@extension kinded projected functions are deprecated", ReplaceWith("combine(arg1)", "arrow.core.combine"))
```
nomisRev, line 24:
```suggestion
@Deprecated("@extension kinded projected functions are deprecated", ReplaceWith("this + arg1"))
```

There is no `combine` defined on `Ordering`

arrow-core/src/main/kotlin/arrow/core/extensions/ordering/semigroup/OrderingSemigroup.kt
```diff
@@ -0,0 +1,46 @@
...
+}
+
+@JvmName("maybeCombine")
+@Suppress(
+  "UNCHECKED_CAST",
+  "USELESS_CAST",
+  "EXTENSION_SHADOWED_BY_MEMBER",
+  "UNUSED_PARAMETER"
+)
+@Deprecated("@extension kinded projected functions are deprecated", ReplaceWith("maybeCombine(arg1)", "arrow.core.combine"))
```
nomisRev, line 36:
```suggestion
@Deprecated("@extension kinded projected functions are deprecated", ReplaceWith("arg1?.plus(this)"))
```

nomisRev (changes requested):
Couple of nit picks and incorrect `@Deprecated`. Thanks @pablisco

pablisco:
> Couple of nit picks and incorrect `@Deprecated`. Thanks @pablisco 

What did I do this time? 😅

franciscodr:
> @1Jajen1 sounds good to keep Ordering but I propose we make one small change to the Order type class.
> `Order` should be a fun interface and the abstract method should be the one for the operator.
> This is so we can construct order instances with no cost using the existing operators that regardless of Order or Ordering will appear concrete in the data type. This operator will be implemented concrete so users can use the syntax `>`, `<=`, ... etc over Option, Either and others that can provide the operator.

@raulraja If we change the abstract method of the Order type class to be the operator, it would be a breaking change, wouldn't it?

arrow-core-data/src/main/kotlin/arrow/core/Ordering.kt
```diff
@@ -23,14 +27,107 @@ sealed class Ordering {
...
 object LT : Ordering()
 object GT : Ordering()
 object EQ : Ordering()
+
+fun Collection<Ordering>.combineAll(): Ordering =
+  if (isEmpty()) OrderingMonoid.empty() else reduce { a, b -> a.combine(b) }
+
+@Suppress("EXTENSION_SHADOWED_BY_MEMBER")
+private object OrderingEq : Eq<Ordering> {
+  override fun Ordering.eqv(b: Ordering): Boolean =
```
franciscodr, line 105: @nomisRev Do we still need to implement these typeclasses for ordering? Or is it enough with having the concrete methods in the Ordering data type?

nomisRev: I think it's still useful to keep these typeclasses, since none of these typeclasses will be deprecated.

nomisRev (approved):
Thanks for the update @franciscodr!

raulraja:
@franciscodr yes that would be a breaking change but the suggestion was not for this PR but for the following release

PR #287, opened 2021-01-11
Move instances to typeclass

nomisRev:
We cannot guarantee that all data types have a `Companion` object, but we can guarantee that a typeclass has a `Companion`. For consistency reason it's therefore most interessting to expose instances on the `Companion` of the typeclass.

This PR moves the new code that exposes the typeclasses to the typeclass Companion.

### DIFF
--- a/arrow-core/src/main/kotlin/arrow/core/extensions/ordering/eq/OrderingEq.kt
+++ b/arrow-core/src/main/kotlin/arrow/core/extensions/ordering/eq/OrderingEq.kt
@@ -30,5 +30,5 @@ fun Ordering.neqv(arg1: Ordering): Boolean = arrow.core.Ordering.eq().run {
   "UNCHECKED_CAST",
   "NOTHING_TO_INLINE"
 )
-@Deprecated("@extension projected functions are deprecated", ReplaceWith("Ordering.eq()"))
+@Deprecated("@extension projected functions are deprecated", ReplaceWith("Order.ordering()", "arrow.core.Order", "arrow.core.ordering"))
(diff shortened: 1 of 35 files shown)


arrow-core/src/main/kotlin/arrow/core/extensions/ordering/eq/OrderingEq.kt
```diff
@@ -30,5 +30,5 @@ fun Ordering.neqv(arg1: Ordering): Boolean = arrow.core.Ordering.eq().run {
   "UNCHECKED_CAST",
   "NOTHING_TO_INLINE"
 )
-@Deprecated("@extension projected functions are deprecated", ReplaceWith("Ordering.eq()"))
+@Deprecated("@extension projected functions are deprecated", ReplaceWith("Order.ordering()", "arrow.core.Order", "arrow.core.ordering"))
```
franciscodr, line 33:
```suggestion
@Deprecated("@extension projected functions are deprecated", ReplaceWith("Eq.ordering()", "arrow.core.Eq", "arrow.core.ordering"))
```

raulraja (approved):
Thanks @nomisRev !

arrow-core/src/main/kotlin/arrow/core/extensions/ordering/eq/OrderingEq.kt
nomisRev: 🙏 Thanks for this catch @franciscodr

franciscodr (approved):
Thanks @nomisRev

PR #306, opened 2021-01-19
Fix incorrect package new instances

nomisRev:
The instances for `Boolean`, `Char`, String` and `Number`s where still in `arrow.core.extensions`, this PR moves them to `arrow.core` instead.


raulraja (approved):
thanks @nomisRev

franciscodr (approved):
Great @nomisRev!

PR #308, opened 2021-01-20
Don't match cancellation in NonFatal

nomisRev:
In Kotlin 1.4 `CancellationException` was added as a cancellation system for coroutines.
We thus shouldn't match `CancellationException` in `NonFatal`, or it would be prone to being swallowed in user code.

I.e. if `fetchUser` is cancelled it throws `CancellationException`, and in the below snippet it could be easily swalled and mapped to a default error type. This would easily cause bugs, and we should let `CancellationException` propagate through the program so it can cancel any outer scopes.

```kotlin
val user = Either.catch {
  fetchUser()
}.mapLeft { DomainError.UnkownError(it) }
```


raulraja (approved):
👌

PR #310, opened 2021-01-21
CU-cn1twz Deprecate @higherkind & @extension for Option 

franciscodr:
- [x] Align
- [x] Alternative
- [x] Applicative
- [x] ApplicativeError
- [x] Apply
- [x] Crosswalk
- [x] Eq (EqK)
- [x] Foldable
- [x] Functor
- [x] FunctorFilter
- [x] Hash
- [x] Monad
- [x] MonadCombine
- [x] MonadError
- [x] MonadFilter
- [x] MonadPlus
- [x] Monoid (MonoidK)
- [x] Monoidal
- [x] Order
- [x] Repeat
- [x] Selective
- [x] Semialign
- [x] Semigroup (SemigroupK)
- [x] Semigroupal
- [x] Show
- [x] Traverse
- [x] TraverseFilter
- [x] Unalign
- [x] Unzip
- [x] Zip


arrow-core/src/main/kotlin/arrow/core/extensions/option/applicative/OptionApplicative.kt
```diff
@@ -70,8 +62,8 @@ fun unit(): Option<Unit> = arrow.core.Option
 @Deprecated(
   "@extension kinded projected functions are deprecated",
   ReplaceWith(
-  "map(arg1)",
-  "arrow.core.map"
+  "fix().map(arg1)",
```
raulraja, line 65: fix is also deprecated and will eventually be removed. I think in this case it would be just replace with `map(arg1)` ?

franciscodr: Yes, the problem is that there is no `map` method defined for a value of type `Kind<ForOption, A>`, but for `Option<A>`. So, if we want to provide a valid replacement, we need to add `fix` in order to have access to the `map` method.

Having said that, I'm ok with removing `fix,` but the user would have to adjust the proposed replacement a bit to have the code compiling.

nomisRev: Yes, that's actually what I've been doing but that might not always lead to a correctly path of deprecating I think. But I'm not sure if it's always possible. For example if a user has done something like `val x = Either.applicative<String>().just(1)` it'll be typed to `EitherOf<String, Int>`.

Now if he called `import arrow.core.extension.functor.map` on it like so `val y = x.map { it + 1 }` it should result in `Either<String, Int>`.

This scenario probably requires a manual rewrite in the assignment of `x` to rewrite it to `val x: Either<String, Int> = 1.right()`. And then he doesn't even need te deprecation tools because now it should've picked the concrete one and the `import arrow.core.extension.functor.map` import should show as unused.
In other cases where we need to switch from the `arrow.core.extensions.*` package to `arrow.core.*` package the deprecation tools should work without `fix()`.

A deprecation path with `fix()` in it would still require a manual rewrite, but it all depends on how the user instantiated the program. Furthermore, people that have written tagless programs today will have to rewrite to concrete ones so deprecation paths for them don't exist. For example, we can't deprecated `Functor#map` or `Applicative#just` with `ReplaceWith`.

franciscodr:
Task linked: [CU-cn1twz Remove extension annotation Option](https://app.clickup.com/t/cn1twz)

arrow-core-data/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -441,8 +616,20 @@ sealed class Option<out A> : OptionOf<A> {
...
+      "this.filterMap(f.andThen { it.orNull() })",
+      "arrow.core.andThen"
+    ),
+    level = DeprecationLevel.WARNING
+  )
   fun <B> filterMap(f: (A) -> Option<B>): Option<B> =
-    flatMap { a -> f(a).fold({ empty<B>() }, { just(it) }) }
+    flatMap(f)
+
+  fun <B> filterMap(f: (A) -> B?): Option<B> =
```
i-walker, line 631: Does it make sense to make this inline, too?

arrow-core/src/main/kotlin/arrow/core/extensions/option/applicativeError/OptionApplicativeError.kt
```diff
@@ -211,10 +199,10 @@ fun <A> ApplicativeError<ForOption, Throwable>.catch(arg1: Function0<A>): Option
 @Deprecated(
   "@extension kinded projected functions are deprecated",
   ReplaceWith(
-  "effectCatch(arg0, arg1)",
-  "arrow.core.Option.effectCatch"
+    "Option.catch(arg0) { arg1() }",
```
danimontoya, line 202: does this function exists for suspend?

danimontoya (commented):
Haven't finished the review yet, will continue after lunch! brb!

arrow-core/src/main/kotlin/arrow/core/extensions/option/apply/OptionApply.kt
```diff
@@ -100,8 +93,8 @@ fun <A, B, Z> Kind<ForOption, A>.map2Eval(
 @Deprecated(
   "@extension kinded projected functions are deprecated",
   ReplaceWith(
-  "map(arg0, arg1, arg2)",
-  "arrow.core.Option.map"
+  "Option.mapN(arg0, arg1) { a, b -> arg2(Tuple2(a, b)) }",
+    "arrow.core.Option", "arrow.core.Tuple2", "arrow.core.mapN"
```
danimontoya, line 97: I think mapN import is not needed
```suggestion
    "arrow.core.Option", "arrow.core.Tuple2"
```

arrow-core/src/main/kotlin/arrow/core/extensions/option/apply/OptionApply.kt
```diff
@@ -123,8 +116,8 @@ fun <A, B, Z> map(
 @Deprecated(
   "@extension kinded projected functions are deprecated",
   ReplaceWith(
-  "mapN(arg0, arg1, arg2)",
-  "arrow.core.Option.mapN"
+    "Option.mapN(arg0, arg1) { a, b -> arg2(Tuple2(a, b)) }",
+    "arrow.core.Option", "arrow.core.Tuple2", "arrow.core.mapN"
```
danimontoya, line 120: same for all the same imports

arrow-core/src/main/kotlin/arrow/core/extensions/option/apply/OptionApply.kt
```diff
@@ -594,8 +587,8 @@ fun <A, B, C, D, E, FF, G, H, I, J, Z> mapN(
 @Deprecated(
   "@extension kinded projected functions are deprecated",
   ReplaceWith(
-  "map2(arg1, arg2)",
-  "arrow.core.map2"
+  "zip(arg1) { a, b -> arg2(Tuple2(a, b)) }",
+    "arrow.core.Tuple2"
```
danimontoya, line 591:
```suggestion
    "arrow.core.Tuple2", "arrow.core.zip"
```

arrow-core/src/main/kotlin/arrow/core/extensions/option/apply/OptionApply.kt
```diff
@@ -614,8 +607,8 @@ fun <A, B, Z> Kind<ForOption, A>.map2(arg1: Kind<ForOption, B>, arg2: Function1<
 @Deprecated(
   "@extension kinded projected functions are deprecated",
   ReplaceWith(
-  "product(arg1)",
-  "arrow.core.product"
+    "zip(arg1) { a, b -> Tuple2(a, b) }",
+    "arrow.core.Tuple2"
```
danimontoya, line 611:
```suggestion
    "arrow.core.Tuple2", "arrow.core.zip"
```

arrow-core/src/main/kotlin/arrow/core/extensions/option/apply/OptionApply.kt
```diff
@@ -634,8 +627,8 @@ fun <A, B> Kind<ForOption, A>.product(arg1: Kind<ForOption, B>): Option<Tuple2<A
 @Deprecated(
   "@extension kinded projected functions are deprecated",
   ReplaceWith(
-  "product(arg1)",
-  "arrow.core.product"
+    "zip(arg1) { tuple, b -> Tuple3(tuple.a, tuple.b, b) }",
+    "arrow.core.Tuple3"
```
danimontoya, line 631: same for the rest
```suggestion
    "arrow.core.Tuple3", "arrow.core.zip"
```

arrow-core/src/main/kotlin/arrow/core/extensions/option/foldable/OptionFoldable.kt
```diff
@@ -296,8 +275,7 @@ fun <A> Kind<ForOption, A>.find(arg1: Function1<A, Boolean>): Option<A> =
 @Deprecated(
   "@extension kinded projected functions are deprecated",
   ReplaceWith(
-  "exists(arg1)",
-  "arrow.core.exists"
+  "exists(arg1)"
```
danimontoya, line 278:
```suggestion
  "exists(arg1)", "arrow.core.exists"
```

arrow-core/src/main/kotlin/arrow/core/extensions/option/foldable/OptionFoldable.kt
```diff
@@ -316,8 +294,7 @@ fun <A> Kind<ForOption, A>.exists(arg1: Function1<A, Boolean>): Boolean =
 @Deprecated(
   "@extension kinded projected functions are deprecated",
   ReplaceWith(
-  "forAll(arg1)",
-  "arrow.core.forAll"
+  "all(arg1)"
```
danimontoya, line 297:
```suggestion
  "all(arg1)", "arrow.core.all"
```

arrow-core/src/main/kotlin/arrow/core/extensions/option/foldable/OptionFoldable.kt
```diff
@@ -336,8 +313,7 @@ fun <A> Kind<ForOption, A>.forAll(arg1: Function1<A, Boolean>): Boolean =
 @Deprecated(
   "@extension kinded projected functions are deprecated",
   ReplaceWith(
-  "all(arg1)",
-  "arrow.core.all"
+  "all(arg1)"
```
danimontoya, line 316:
```suggestion
  "all(arg1)", "arrow.core.all"
```

arrow-core/src/main/kotlin/arrow/core/extensions/option/foldable/OptionFoldable.kt
```diff
@@ -356,8 +332,7 @@ fun <A> Kind<ForOption, A>.all(arg1: Function1<A, Boolean>): Boolean =
 @Deprecated(
   "@extension kinded projected functions are deprecated",
   ReplaceWith(
-  "isEmpty()",
-  "arrow.core.isEmpty"
+  "isEmpty()"
```
danimontoya, line 335: same for missing imports

arrow-core/src/main/kotlin/arrow/core/extensions/option/monadPlus/OptionMonadPlus.kt
```diff
@@ -46,8 +42,8 @@ fun <A> zeroM(): Option<A> = arrow.core.Option
 @Deprecated(
   "@extension kinded projected functions are deprecated",
   ReplaceWith(
-  "plusM(arg1)",
-  "arrow.core.plusM"
+  "orElse { arg1 }",
+  "arrow.core.Option", "arrow.core.orElse"
```
danimontoya, line 46:
```suggestion
  "getOrElse { arg1 }",
  "arrow.core.Option", "arrow.core.getOrElse"
```

arrow-core-data/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -570,15 +994,15 @@ data class Some<out T>(val t: T) : Option<T>() {
...
+inline fun <T> Option<T>.getOrElse(default: () -> T): T = fold({ default() }, ::identity)
 
 /**
  * Returns this option's if the option is nonempty, otherwise
  * returns another option provided lazily by `default`.
  *
  * @param alternative the default option if this is empty.
  */
-fun <A> OptionOf<A>.orElse(alternative: () -> Option<A>): Option<A> = if (fix().isEmpty()) alternative() else fix()
+inline fun <A> OptionOf<A>.orElse(alternative: () -> Option<A>): Option<A> = if (fix().isEmpty()) alternative() else fix()
```
danimontoya, line 1005: this one should be deprecated as well??

arrow-core-data/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -570,15 +994,15 @@ data class Some<out T>(val t: T) : Option<T>() {
...
 
 infix fun <T> OptionOf<T>.or(value: Option<T>): Option<T> = if (fix().isEmpty()) {
```
danimontoya, line 1007: this one too?

danimontoya (commented):
Great work paco!! thanks for the big effor, Option is huge!! 🚀 🙌  
only added a few things related to imports and some suggestions like changing orElse for getOrElse which is closer to Kotlin std! Still, would be nice if @nomisRev takes another look 😄

danimontoya:
Oh one more thing.. `@extension` should be removed from `arrow.core.extensions.option`

arrow-core-data/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -628,7 +628,7 @@ sealed class Option<out A> : OptionOf<A> {
   fun <B> filterMap(f: (A) -> Option<B>): Option<B> =
     flatMap(f)
 
-  fun <B> filterMap(f: (A) -> B?): Option<B> =
+  inline fun <B> filterMap(f: (A) -> B?): Option<B> =
```
nomisRev: Duplicated with `mapNotNull` should we keep one or the other? I think on `Iterable<A>`/Collections we've deprecated `filterMap` in favor of `mapNotNull` which for those types comes from Kotlin's Std.

arrow-core-data/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -395,7 +418,137 @@ sealed class Option<out A> : OptionOf<A> {
...
+    inline fun <A> catch(recover: (Throwable) -> Unit, f: () -> A): Option<A> =
+      try {
+        just(f())
+      } catch (t: Throwable) {
+        raiseError(recover(t.nonFatalOrThrow()))
+      }
+
     fun <A> empty(): Option<A> = None
+
+    fun <A> raiseError(e: Unit): Option<A> = None
```
nomisRev, line 430: I don't we should add this API for `Option` and otherwise I'd define it as `fun <A> raiseError(): Option<A> = None`.

IMO `just`, `raiseError` are counter intinuitive if not used in a polymorphic context. I'm unsure if `Option.just(1)` and `Either.just(1)` bring any additional value over `Option.some(1)` or `Either.right(1)`?

We need to review all these different constructors in all the data types before 0.12.0. I think for `Either` we have `Either.just(1)`, `Either.right(1)`, `1.right()`, `Either.raiseError("1")`, `Either.left("1")` and `"1".left()"`.
Futhermore Kotlin favors a type argument over returning `Nothing` like we currently do with `right()` and `left()` in `Either`. For example Kotlin Std: `emptyList<A>()` instead of `emptyList()` which would return `List<Nothing>`.

arrow-core-data/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -441,8 +616,20 @@ sealed class Option<out A> : OptionOf<A> {
...
+      "this.filterMap(f.andThen { it.orNull() })",
+      "arrow.core.andThen"
+    ),
+    level = DeprecationLevel.WARNING
+  )
   fun <B> filterMap(f: (A) -> Option<B>): Option<B> =
-    flatMap { a -> f(a).fold({ empty<B>() }, { just(it) }) }
+    flatMap(f)
+
+  inline fun <B> filterMap(f: (A) -> B?): Option<B> =
```
nomisRev, line 631: Duplicate with `mapNonNull`, should me favor `mapNotNull` instead? We do that for `Iterable<A>` & Collections where we favor `kotlin.collections.mapNotNull` over providing our own `filterMap` function.

arrow-core-data/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -495,47 +712,254 @@ sealed class Option<out A> : OptionOf<A> {
...
+   * val none: Option<Int> = None
+   * none.exists { it > 10 }      // Result: false
+   * ```
+   *
    * @param predicate the predicate to test
    */
-  fun exists(predicate: Predicate<A>): Boolean = fold({ false }, { a -> predicate(a) })
+  inline fun exists(predicate: (A) -> Boolean): Boolean = fold({ false }, predicate)
+
+  inline fun filterA(predicate: (A) -> Iterable<Boolean>): Iterable<Option<A>> =
```
nomisRev, line 736: `filterA`? Where does this name come from 🤔?

arrow-core-data/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -495,47 +712,254 @@ sealed class Option<out A> : OptionOf<A> {
...
+    traverseFilterValidated { a -> predicate(a).map { if (it) Some(a) else None } }
 
   /**
    * Returns true if this option is empty '''or''' the predicate
    * $p returns true when applied to this $option's value.
    *
    * @param p the predicate to test
    */
-  fun forall(p: Predicate<A>): Boolean = fold({ true }, p)
+  inline fun forall(p: (A) -> Boolean): Boolean = fold({ true }, p)
```
nomisRev, line 751: This is named `all` in Kotlin, https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/all.html.
We should deprecate towards `all`.

arrow-core-data/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -614,10 +1043,317 @@ fun <T> Iterable<T>.lastOrNone(predicate: (T) -> Boolean): Option<T> = this.last
...
+}
+
+inline fun <A> Option<A>.handleError(f: (Unit) -> A): Option<A> =
+  handleErrorWith { Option.just(f(Unit)) }
+
+inline fun <A> Option<A>.handleErrorWith(f: (Unit) -> Option<A>): Option<A> =
+  if (isEmpty()) f(Unit) else this
+
+inline fun <reified B> Option<*>.traverseFilterIsInstance(): List<Option<B>> =
+  filterA { a -> listOf(a is B) }.map { it.map { a -> a as B } }
```
nomisRev, line 1093:
```suggestion
  fold({ emptyList<Option<B>>() }, { a -> (a as? B)?.let(::listOf) ?: emptyList() })
```

arrow-core-data/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -614,10 +1043,317 @@ fun <T> Iterable<T>.lastOrNone(predicate: (T) -> Boolean): Option<T> = this.last
...
+
+inline fun <E, reified B> Option<*>.traverseFilterIsInstanceEither(): Either<E, Option<B>> =
+  filterAEither { a -> Right(a is B) }.map { it.map { a -> a as B } }
```
nomisRev, line 1096: How does this function work? 🤔 This will result in a `ClassCastException` when `a` is not `B`.

arrow-core-data/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -614,10 +1043,317 @@ fun <T> Iterable<T>.lastOrNone(predicate: (T) -> Boolean): Option<T> = this.last
...
+
+inline fun <E, reified B> Option<*>.traverseFilterIsInstanceValidated(): Validated<E, Option<B>> =
+  filterAValidated { a -> Valid(a is B) }.map { it.map { a -> a as B } }
```
nomisRev, line 1099: Idem `traverseFilterIsInstanceEither`

arrow-core-data/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -614,10 +1043,317 @@ fun <T> Iterable<T>.lastOrNone(predicate: (T) -> Boolean): Option<T> = this.last
...
+ *   "Hello".some().map({ "$it World" }).widen()
+ *   //sampleEnd
+ *   println(result)
+ *  }
+ *  ```
+ */
+fun <B, A : B> Option<A>.widen(): Option<B> =
+  this
+
+fun <A> Option<A>.eqv(EQA: Eq<A>, other: Option<A>): Boolean = when (this) {
```
nomisRev, line 1237: These will not be added for the data types. Still need to create a clean-up ticket for the already done types.

arrow-core/src/main/kotlin/arrow/core/extensions/option/alternative/OptionAlternative.kt
```diff
@@ -69,8 +54,8 @@ fun <A> Kind<ForOption, A>.many(): Option<SequenceK<A>> = arrow.core.Option.alte
 @Deprecated(
   "@extension kinded projected functions are deprecated",
   ReplaceWith(
-  "alt(arg1)",
-  "arrow.core.alt"
+  "orElse { arg1 }",
+  "arrow.core.orElse"
```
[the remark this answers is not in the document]
nomisRev: `orElse` and `getOrElse` do 2 different things.

`optOne.orElse { optTwo }` results in `optTwo` when `optOne == None`, while `getOrElse` will extract the value. `optOne.getOrElse { 2 }` will return `2` if `optOne == None`.

`orElse` is correct here.

arrow-core/src/main/kotlin/arrow/core/extensions/option/alternative/OptionAlternative.kt
```diff
@@ -89,8 +74,8 @@ infix fun <A> Kind<ForOption, A>.alt(arg1: Kind<ForOption, A>): Option<A> =
 @Deprecated(
   "@extension kinded projected functions are deprecated",
   ReplaceWith(
-  "orElse(arg1)",
-  "arrow.core.orElse"
+    "orElse { arg1 }",
+    "arrow.core.orElse"
```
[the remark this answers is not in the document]
nomisRev: Idem, `orElse` is correct here.

arrow-core/src/main/kotlin/arrow/core/extensions/option/alternative/OptionAlternative.kt
```diff
@@ -109,8 +94,8 @@ fun <A> Kind<ForOption, A>.orElse(arg1: Kind<ForOption, A>): Option<A> =
 @Deprecated(
   "@extension kinded projected functions are deprecated",
   ReplaceWith(
-  "combineK(arg1)",
-  "arrow.core.combineK"
+    "orElse { arg1 }",
+    "arrow.core.orElse"
```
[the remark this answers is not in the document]
nomisRev: Same 😁

arrow-core/src/main/kotlin/arrow/core/extensions/option/alternative/OptionAlternative.kt
```diff
@@ -129,8 +114,8 @@ fun <A> Kind<ForOption, A>.combineK(arg1: Kind<ForOption, A>): Option<A> =
 @Deprecated(
   "@extension kinded projected functions are deprecated",
   ReplaceWith(
-  "optional()",
-  "arrow.core.optional"
+  "map<Option<A>>(::Some).orElse { Option.just(None) }",
+  "arrow.core.None", "arrow.core.Option", "arrow.core.Some", "arrow.core.orElse"
```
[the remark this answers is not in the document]
nomisRev: Same 😅

arrow-core/src/main/kotlin/arrow/core/extensions/option/alternative/OptionAlternative.kt
```diff
@@ -167,8 +152,8 @@ fun guard(arg0: Boolean): Option<Unit> = arrow.core.Option
 @Deprecated(
   "@extension kinded projected functions are deprecated",
   ReplaceWith(
-  "lazyOrElse(arg1)",
-  "arrow.core.lazyOrElse"
+    "orElse { arg1() }",
+    "arrow.core.orElse"
```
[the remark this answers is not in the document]
nomisRev: Same 😅

arrow-core/src/main/kotlin/arrow/core/extensions/option/applicativeError/OptionApplicativeError.kt
nomisRev: The function is `inline` so you can call `suspend` in it's lambda.

arrow-core/src/main/kotlin/arrow/core/extensions/option/monadPlus/OptionMonadPlus.kt
nomisRev: same `orElse` is the correct method here.

nomisRev (commented):
Looking good. Couple suggestions and questions. Amazing work this was a data type with a lot of instances 👏 👏 👏

arrow-core-data/src/main/kotlin/arrow/core/Option.kt
raulraja: I think we should get rid of just and raiseError. people will choose constructors over just and agreed that Option should not have a raiseError. For either extensions like .right(), .left() etc I think we should consolidate to use constructors or constructor like functions outside of the companion. I don't see any value to the dot syntax.or the companion factories beside confusing users with multiple ways to do the same.

arrow-core-data/src/main/kotlin/arrow/core/Option.kt
raulraja: I believe so, we should use mapNotNull instead of filterMap across the board

arrow-core-data/src/main/kotlin/arrow/core/Option.kt
nomisRev: 100% agreed.

franciscodr: Makes sense, adding `raiseError` sounded weird to me but I put it because Option implemented `ApplicativeError` so I wasn't sure if we're ok removing that method.

arrow-core-data/src/main/kotlin/arrow/core/Option.kt
franciscodr: Ok, I'll do that.

arrow-core-data/src/main/kotlin/arrow/core/Option.kt
franciscodr: I guess `A` comes from `Applicative` - [see here](fadb9a31ffb0a4b8f9bc88c674cefae2680a3b415cf7c5d2900aa7cd9f74afe0L77). I will rename it to `filterIterable`

nomisRev: Yes, `filterIterable` would be great 👌

nomisRev (approved):
This is looking good to me! Great work @franciscodr!! One more of the data types with the most instances tackled 👏 👏 👏

arrow-core/src/main/kotlin/arrow/core/extensions/option.kt
```diff
@@ -82,24 +78,20 @@ interface OptionSemigroup<A> : Semigroup<Option<A>> {
...
 }
 
-@extension
 interface OptionMonoid<A> : Monoid<Option<A>>, OptionSemigroup<A> {
   override fun SG(): Semigroup<A>
   override fun empty(): Option<A> = None
 }
 
-@extension
 interface OptionApplicativeError : ApplicativeError<ForOption, Unit>, OptionApplicative {
```
nomisRev, line 95:
```
@Deprecated(
  message = "ApplicativeError typeclass is deprecated and will be removed in 0.13.0. Use concrete methods on Const",
  level = DeprecationLevel.WARNING
)
```

We should add this for all the deprecated interface instances. We need to do this retroactively for the already done PRs, I think we forgot to do it there too.

arrow-core/src/main/kotlin/arrow/core/extensions/option.kt
```diff
@@ -67,7 +64,6 @@ import arrow.core.extensions.traverse as optionTraverse
 import arrow.core.extensions.traverseFilter as optionTraverseFilter
 import arrow.core.select as optionSelect
 
-@extension
 interface OptionSemigroup<A> : Semigroup<Option<A>> {
```
nomisRev, line 67:
```
@Deprecated(
  "Typeclass instance have been moved to the companion object of the typeclass.",
  ReplaceWith("Semigroup.option()", "arrow.core.option", "arrow.typeclasses.Semigroup"),
  DeprecationLevel.WARNING
)
```

danimontoya (approved):
Well done paco, thanks!! 🙌 🚀

PR #317, opened 2021-01-26
CU-cn28ya Deprecate @higherkind & @extension for Ior 

franciscodr:
- [x] Applicative
- [x] Apply
- [x] Bicrosswalk
- [x] Bifoldable
- [x] Bifunctor
- [x] Bitraverse
- [x] Crosswalk
- [x] Eq (EqK, EqK2)
- [x] Foldable
- [x] Functor
- [x] Hash
- [x] Monad
- [x] Order
- [x] Semigroup
- [x] Show
- [x] Traverse

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/core/Ior.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Ior.kt
@@ -631,6 +631,27 @@ sealed class Ior<out A, out B> : IorOf<A, B> {
       { a, b -> Validated.mapN(SA, fa(a), fb(b)) { aa, c -> Both(aa, c) } }
     )
 
+  inline fun <C> crosswalk(fa: (B) -> Iterable<C>): List<Ior<A, C>> =
+    fold(
+      { emptyList() },
+      { b -> fa(b).map { Right(it) } },
+      { a, b -> fa(b).map { Both(a, it) }}
+    )
+
+  inline fun <K, V> crosswalkMap(fa: (B) -> Map<K, V>): Map<K, Ior<A, V>> =
+    fold(
+      { emptyMap() },
+      { b -> fa(b).mapValues { Right(it.value) } },
+      { a, b -> fa(b).mapValues { Both(a, it.value) }}
+    )
+
+  inline fun <A, B, C> crosswalkNull(ior: Ior<A, B>, fa: (B) -> C?): Ior<A, C>? =
+    ior.fold(
+      { a -> Left(a) },
+      { b -> fa(b)?.let { Right(it) } },
+      { a, b -> fa(b)?.let { Both(a, it) }}
+    )
+
   /**
    *  Applies [f] to an [B] inside [Ior] and returns the [Ior] structure with a pair of the [B] value and the
    *  computed [C] value as result of applying [f]
--- a/arrow-core-data/src/main/kotlin/arrow/core/Ior.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Ior.kt
@@ -2,6 +2,7 @@ package arrow.core
 
 import arrow.Kind
 import arrow.typeclasses.Applicative
+import arrow.typeclasses.Monoid
 import arrow.typeclasses.Semigroup
 import arrow.typeclasses.Show
 
@@ -144,6 +145,10 @@ sealed class Ior<out A, out B> : IorOf<A, B> {
     fun <A, B> leftNel(a: A): IorNel<A, B> = Left(NonEmptyList.of(a))
 
     fun <A, B> bothNel(a: A, b: B): IorNel<A, B> = Both(NonEmptyList.of(a), b)
+
+    val unit: Ior<Nothing, Unit> = Right(Unit)
+
+    fun <L> unit(): Ior<L, Unit> = unit
   }
 
   /**
@@ -495,6 +500,28 @@ fun <A, B> Tuple2<A, B>.bothIor(): Ior<A, B> = Ior.Both(this.a, this.b)
 fun <A> A.leftIor(): Ior<A, Nothing> = Ior.Left(this)
 
 fun <A> A.rightIor(): Ior<Nothing, A> = Ior.Right(this)
+
+fun <A, B> Ior<A, B>.replicate(SA: Semigroup<A>, n: Int): Ior<A, List<B>> =
+  if (n <= 0) Ior.Right(emptyList())
+  else when (this) {
+    is Ior.Right -> Ior.Right(List(n) { value })
+    is Ior.Left -> this
+    is Ior.Both -> bimap(
+      { List(n - 1) { leftValue}.fold(leftValue, { acc, a -> SA.run { acc + a }}) },
+      { List(n) { rightValue } }
+    )
+  }
+
+fun <A, B> Ior<A, B>.replicate(SA: Semigroup<A>, n: Int, MB: Monoid<B>): Ior<A, B> =
+  if (n <= 0) Ior.Right(MB.empty())
+  else when (this) {
+    is Ior.Right -> Ior.Right(MB.run { List(n) { value }.combineAll() })
+    is Ior.Left -> this
+    is Ior.Both -> bimap(
+      { List(n - 1) { leftValue}.fold(leftValue, { acc, a -> SA.run { acc + a }}) },
+      { MB.run { List(n) { rightValue }.combineAll() } }
+    )
+  }
 
 fun <E, A, B, Z> Ior<E, A>.zip(SE: Semigroup<E>, fb: Ior<E, B>, f: (A, B) -> Z): Ior<E, Z> =
   ap(SE, fb.map { b: B -> { a: A -> f(a, b) } })
--- a/arrow-core-data/src/main/kotlin/arrow/core/Ior.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Ior.kt
@@ -744,6 +744,9 @@ fun <A, B> Ior<A, B>.eqv(EQA: Eq<A>, EQB: Eq<B>, b: Ior<A, B>): Boolean = when (
   }
 }
 
+fun <A, B> Ior<A, B>.neqv(EQA: Eq<A>, EQB: Eq<B>, b: Ior<A, B>): Boolean =
+  !eqv(EQA, EQB, b)
+
 @Suppress("NOTHING_TO_INLINE")
 inline fun <A, B> Ior<A, Ior<A, B>>.flatten(SA: Semigroup<A>): Ior<A, B> =
   flatMap(SA, ::identity)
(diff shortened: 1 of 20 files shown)


franciscodr:
Task linked: [CU-cn28ya Remove extension annotation Ior](https://app.clickup.com/t/cn28ya)

arrow-core-data/src/main/kotlin/arrow/core/Ior.kt
```diff
@@ -486,8 +904,247 @@ fun <A> A.leftIor(): Ior<A, Nothing> = Ior.Left(this)
...
+  compare(OA, OB, b) != LT
+
+fun <A, B> Ior<A, B>.max(OA: Order<A>, OB: Order<B>, b: Ior<A, B>): Ior<A, B> =
+  if (gt(OA, OB, b)) this else b
+
+fun <A, B> Ior<A, B>.min(OA: Order<A>, OB: Order<B>, b: Ior<A, B>): Ior<A, B> =
+  if (lt(OA, OB, b)) this else b
+
+fun <A, B> Ior<A, B>.sort(OA: Order<A>, OB: Order<B>, b: Ior<A, B>): Pair<Ior<A, B>, Ior<A, B>> =
+  if (gte(OA, OB, b)) this to b else b to this
```
nomisRev, line 976: I think we don't want to add all these methods. Might be similar for `Show`, `eqv` & `hash`.

arrow-core-data/src/main/kotlin/arrow/core/Ior.kt
```diff
@@ -134,6 +149,161 @@ sealed class Ior<out A, out B> : IorOf<A, B> {
...
+      c: Ior<A, C>,
+      d: Ior<A, D>,
+      e: Ior<A, E>,
+      f: Ior<A, F>,
+      g: Ior<A, G>,
+      h: Ior<A, H>,
+      i: Ior<A, I>,
+      j: Ior<A, J>,
+      k: Ior<A, K>,
+      crossinline map: (B, C, D, E, F, G, H, I, J, K) -> L
```
nomisRev, line 280: Actually, I think with this implementation we still can't call `suspend` functions inside can we?

arrow-core-data/src/main/kotlin/arrow/core/Ior.kt
```diff
@@ -472,6 +887,9 @@ inline fun <A, B, D> Ior<A, B>.flatMap(SG: Semigroup<A>, f: (B) -> Ior<A, D>): I
 fun <A, B, D> Ior<A, B>.ap(SG: Semigroup<A>, ff: IorOf<A, (B) -> D>): Ior<A, D> =
   flatMap(SG) { a -> ff.fix().map { f -> f(a) } }
 
+fun <A, B, D> Ior<A, B>.ap(SG: Semigroup<A>, ff: Eval<Ior<A, (B) -> D>>): Eval<Ior<A, D>> =
```
nomisRev, line 890:
```suggestion
fun <A, B, D> Ior<A, B>.apEval(SG: Semigroup<A>, ff: Eval<Ior<A, (B) -> D>>): Eval<Ior<A, D>> =
```

arrow-core/src/main/kotlin/arrow/core/extensions/ior/eqK2/IorEqK2.kt
```diff
@@ -25,12 +20,8 @@ internal val eqK2_singleton: IorEqK2 = object : arrow.core.extensions.IorEqK2 {}
   "UNUSED_PARAMETER"
 )
 @Deprecated(
-  "@extension kinded projected functions are deprecated",
-  ReplaceWith(
-  "eqK(arg1, arg2, arg3)",
-  "arrow.core.eqK"
-  ),
-  DeprecationLevel.WARNING
+  "Kind/type constructors will be deprecated, so this typeclass will no longer be available from 0.13.0",
```
nomisRev, line 23: This can be deprecated with `eqv`.

nomisRev (commented):
Looks great overall. Couple of discussion points.

arrow-core-data/src/main/kotlin/arrow/core/Ior.kt
franciscodr: You're right, I will remove those methods and the instances of `Show`, `Eq`, `Order` and `Hash` for `Ior` datatype. In that case, are we just deprecating those typeclasses with a message saying they will be removed in version 0.13?

nomisRev: Yes. I'm not sure if we need to elaborate anywhere why we're doing this for futher reference. Almost feels like the 0.13.0 release needs to be accompanied by a couple of blogposts to explain our rationale about Arrow's progression 😅

arrow-core-data/src/main/kotlin/arrow/core/Ior.kt
```diff
@@ -277,28 +272,42 @@ sealed class Ior<out A, out B> : IorOf<A, B> {
...
-                          map(bb, cc, dd, ee, ff, gg, hh, ii, jj, kk)
-                        }
-                      }
-                    }
-                  }
-                }
-              }
-            }
-          }
+      if (b is Right && c is Right && d is Right && e is Right && f is Right && g is Right && h is Right && i is Right && j is Right && k is Right) {
```
nomisRev, line 277: Awesome! This should be much faster than the derived implementation from `Apply`, and this bytecode should also be much shorter than the `flatMap` derived version.

arrow-core-data/src/main/kotlin/arrow/core/Ior.kt
```diff
@@ -430,18 +621,225 @@ sealed class Ior<out A, out B> : IorOf<A, B> {
     override val isLeft: Boolean get() = false
     override val isBoth: Boolean get() = true
 
-    override fun toString(): String = show(Show.any(), Show.any())
+    override fun toString(): String = "Both(${ leftValue.toString() }, ${ rightValue.toString() })"
   }
 
-  fun show(SL: Show<A>, SR: Show<B>): String = fold(
```
nomisRev, line 436: Needs to be `@Deprecated`.

arrow-core/src/main/kotlin/arrow/core/extensions/ior.kt
```diff
@@ -226,7 +226,15 @@ interface IorEqK2 : EqK2<ForIor> {
 interface IorShow<L, R> : Show<Ior<L, R>> {
   fun SL(): Show<L>
   fun SR(): Show<R>
-  override fun Ior<L, R>.show(): String = show(SL(), SR())
```
nomisRev, line 229: If deprecated this can stay.

arrow-core-data/src/main/kotlin/arrow/core/Ior.kt
franciscodr: That's true! Good catch 😄

arrow-core-data/src/main/kotlin/arrow/core/Ior.kt
```diff
@@ -492,8 +909,181 @@ fun <A> A.leftIor(): Ior<A, Nothing> = Ior.Left(this)
...
+) : Semigroup<Ior<A, B>> {
+
+  override fun Ior<A, B>.combine(b: Ior<A, B>): Ior<A, B> =
+    combine(SGA, SGB, b)
+
+  override fun Ior<A, B>.maybeCombine(b: Ior<A, B>?): Ior<A, B> =
+    b?.let { combine(SGA, SGB, it) } ?: this
+}
+
+operator fun <A : Comparable<A>, B : Comparable<B>> Ior<A, B>.compareTo(other: Ior<A, B>): Int = fold(
```
nomisRev, line 1080: This cannot be implemented inside `Ior` since we cannot constraint `A` and `B` to `Comparable` there? So we can also not implemented `Comparable` for `Ior<A, B>`, correct?

arrow-core/src/main/kotlin/arrow/core/extensions/ior.kt
```diff
@@ -72,19 +70,16 @@ interface IorSemigroup<L, R> : Semigroup<Ior<L, R>> {
     }
 }
 
-@extension
 @undocumented
 interface IorFunctor<L> : Functor<IorPartialOf<L>> {
```
nomisRev, line 74:
```
@Deprecated(
  message = "Functor typeclass is deprecated and will be removed in 0.13.0. Use concrete methods on Const",
  level = DeprecationLevel.WARNING
)
```

Typeclasses that are going to be removed/deprecated should be annotated with the above deprecation

arrow-core/src/main/kotlin/arrow/core/extensions/ior.kt
```diff
@@ -42,7 +41,6 @@ import arrow.typeclasses.Traverse
 import arrow.typeclasses.hashWithSalt
 import arrow.undocumented
 
-@extension
 interface IorSemigroup<L, R> : Semigroup<Ior<L, R>> {
```
nomisRev, line 44:
```
@Deprecated(
  "Typeclass instance have been moved to the companion object of the typeclass.",
  ReplaceWith("Semigroup.ior()", "arrow.core.ior", "arrow.typeclasses.Semigroup"),
  DeprecationLevel.WARNING
)
```

Typeclasses that are not getting deprecated should be annoted with the above deprecation

nomisRev (commented):
Looks great! 👏 👏 
One question, and some missing deprcation on the deprcated instances.

arrow-core-data/src/main/kotlin/arrow/core/Ior.kt
raulraja: makes sense

franciscodr: Yes, as we're adding an additional constraint for `A` and `B,` we cannot implement `compareTo` inside `Ior`. With this approach, `compareTo` will be available for `Ior` if and only if `A` and `B` implement `Comparable`. Does it make sense?

nomisRev (approved):
Looks great @franciscodr! 👏 👏 We're getting close 🥳

arrow-core/src/main/kotlin/arrow/core/extensions/ior/apply/IorApply.kt
```diff
@@ -75,8 +71,8 @@ fun <L, A, B> Kind<Kind<ForIor, L>, A>.apEval(
 @Deprecated(
   "@extension kinded projected functions are deprecated",
   ReplaceWith(
-  "map2Eval(SL, arg1, arg2)",
-  "arrow.core.map2Eval"
+  "this.zipEval<L, A, B, Z>(SL, arg1) { a, b -> arg2(Tuple2(a, b)) }",
+  "arrow.core.zipEval"
```
danimontoya, line 75:
```suggestion
  "arrow.core.zipEval", "arrow.core.Tuple2"
```

danimontoya (approved):
Thank you @franciscodr !! 🙌 🚀

PR #318, opened 2021-02-01
Remove kapt from core-data

danimontoya:
- Remove kapt from core-data
- Remove higherkind annotations
- Add previously generated code to each typeclasses

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/core/SortedMapK.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/SortedMapK.kt
@@ -1,12 +1,18 @@
 package arrow.core
 
 import arrow.Kind
-import arrow.higherkind
 import arrow.typeclasses.Applicative
 import arrow.typeclasses.Show
 import kotlin.collections.flatMap
 
-@higherkind
+class ForSortedMapK private constructor() { companion object }
+typealias SortedMapKOf<A, B> = arrow.Kind2<ForSortedMapK, A, B>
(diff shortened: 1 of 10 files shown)


arrow-core-data/src/main/kotlin/arrow/core/SortedMapK.kt
```diff
@@ -1,12 +1,18 @@
 package arrow.core
 
 import arrow.Kind
-import arrow.higherkind
 import arrow.typeclasses.Applicative
 import arrow.typeclasses.Show
 import kotlin.collections.flatMap
 
-@higherkind
+class ForSortedMapK private constructor() { companion object }
```
raulraja, line 8: All these should be deprecated without replacement now that they are spelled out in the sources.

raulraja (approved):
We need to deprecate those types but otherwise looks good, thanks @danimontoya !

arrow-core-data/src/main/kotlin/arrow/core/SortedMapK.kt
danimontoya: yes!! there are already ongoing PRs for Const, Eval, Ior deprecating them.. so they are being deprecated for each typeclass separately.. but in order to remove kapt and unblock one issue I have, this PR removing kapt is needed. Deprecation of all those will come soon too!

rachelcarmena (approved):
:raised_hands:

nomisRev (approved):

PR #320, opened 2021-02-03
CU-cn1pxw Deprecate AndThen and always use underneath `andThen` and `compose`.

nomisRev:
This PR makes `AndThen` internal, and uses it under `andThen` and `compose` to automatically try and make them stack-safe.

The rationale of deprecating `AndThen` to make it `internal` is that `AndThen` cannot _guarantee_ stack-safety anyway. It makes the assumption that `val f: (A) -> B` is of depth `0`.
If `f` is not of depth `0`, it can still result in a stack-overflow.

`andThen` and `compose` are duplicated between `arrow-syntax`, and `arrow-core` but we might merge these two modules in the `0.13.0` release, at that point we can deprecate the `andThen` and `compose` in `arrow.core` in favor of `arrow.syntax.*` or the other way around.

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/core/AndThen.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/AndThen.kt
@@ -1,13 +1,41 @@
 package arrow.core
 
+@Deprecated(
+  "Kind is deprecated, and will be removed in 0.13.0. Please use one of the provided concrete methods instead",
+  level = DeprecationLevel.WARNING
+)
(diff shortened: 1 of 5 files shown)


franciscodr:
Task linked: [CU-cn1pxw Remove extension annotation for AndThen](https://app.clickup.com/t/cn1pxw)

raulraja (approved):
Looks great @nomisRev ! thanks! 🙏

danimontoya (approved):
Great @nomisRev !! 🚀 🚀

PR #321, opened 2021-02-03
Deprecate Show

nomisRev:
This PR deprecates `Show` since Kotlin, and Kotlin's Std doesn't take `Show` into account nor does Kotlin have the power to automatically derive or inject `Show` anywhere. This makes `Show` very cumbersome to work with, see [Tuple10Show.kt](https://github.com/arrow-kt/arrow-core/compare/sv-deprecate-show?expand=1#diff-7f26da547470ab066e076a5795cdd1e670a17d5eded730b20bf9a7ecd844be86).

The use-cases of `Show` seemed very limited and niche to us, so we've decided to exclude it from Arrow in 1.0.0. The common technique for this is to use `data class`, which automatically implements `toString()` and correctly implementing `toString()`.

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
@@ -1085,6 +1085,7 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
     }
   }
 
+  @Deprecated(ShowDeprecation)
   fun show(SL: Show<A>, SR: Show<B>): String = fold(
     { "Left(${SL.run { it.show() }})" },
     { "Right(${SR.run { it.show() }})" }
--- a/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
@@ -1063,7 +1063,7 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
     override val isLeft = true
     override val isRight = false
 
-    override fun toString(): String = show(Show.any(), Show.any())
+    override fun toString(): String = "Either.Left($a)"
 
     companion object {
       operator fun <A> invoke(a: A): Either<A, Nothing> = Left(a)
@@ -1078,7 +1078,7 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
     override val isLeft = false
     override val isRight = true
 
-    override fun toString(): String = show(Show.any(), Show.any())
+    override fun toString(): String = "Either.Right($b)"
 
     companion object {
       operator fun <B> invoke(b: B): Either<Nothing, B> = Right(b)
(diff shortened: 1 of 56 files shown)


arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -1078,20 +1078,21 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
...
 
-    override fun toString(): String = show(Show.any(), Show.any())
+    override fun toString(): String = "Either.Right($b)"
 
     companion object {
       operator fun <B> invoke(b: B): Either<Nothing, B> = Right(b)
     }
   }
 
   fun show(SL: Show<A>, SR: Show<B>): String = fold(
```
franciscodr, line 1088:
```suggestion
  @Deprecated(ShowDeprecation)
  fun show(SL: Show<A>, SR: Show<B>): String = fold(
```

arrow-core-data/src/main/kotlin/arrow/core/Eval.kt
```diff
@@ -532,6 +544,9 @@ sealed class Eval<out A> : EvalOf<A> {
     abstract fun <S> run(s: S): Eval<A>
     override fun memoize(): Eval<A> = Memoize(this)
     override fun value(): A = evaluate(this)
+
+    override fun toString(): String =
+      "Eval.FlatMao(..)"
```
franciscodr, line 549:
```suggestion
      "Eval.FlatMap(..)"
```

arrow-core-data/src/main/kotlin/arrow/core/Eval.kt
```diff
@@ -547,7 +562,13 @@ sealed class Eval<out A> : EvalOf<A> {
     override fun value(): A = result.getOrElse {
       evaluate(eval).also { result = Some(it) }
     }
+
+    override fun toString(): String =
+      "Eval.Memoize($eval)"
   }
+
+  override fun toString(): String =
```
franciscodr, line 570: I was wondering why we need the `toString` method in the sealed class

nomisRev: I guess it's not needed if it's defined for all the cases.

franciscodr:
I think we need to remove the instance of Show for Validated and its implementation:
- [This line](https://github.com/arrow-kt/arrow-core/pull/321/files#diff-458531809e0125ff24d909bd1213c7e01ef68ecdbbff1537a1a13e4e163cbf3cR961)
- And [this line](https://github.com/arrow-kt/arrow-core/pull/321/files#diff-458531809e0125ff24d909bd1213c7e01ef68ecdbbff1537a1a13e4e163cbf3cR1270)

franciscodr (approved):
Thanks, @nomisRev!

i-walker (approved):
Thanks @nomisRev Looks great

PR #322, opened 2021-02-03
Deprecate Eq

nomisRev:
This PR deprecates `Eq` since Kotlin, and Kotlin's Std doesn't take `Eq` into account nor does Kotlin have the power to automatically derive or inject `Eq` anywhere. This makes `Eq` very cumbersome to work with, see [`Tuple10Eq.kt`](https://github.com/arrow-kt/arrow-core/compare/sv-deprecate-eq?expand=1#diff-71c97dda0b6a770b17e372fadc91f43578fe2e833d77959444cab6885fc000b8)

The use-cases of `Eq` seemed very limited and niche to us, so we've decided to exclude it from Arrow in 1.0.0. The common technique for this is to use data class, which automatically implements `equals()`.

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
@@ -4,7 +4,6 @@ import arrow.Kind
 import arrow.core.Either.Companion.resolve
 import arrow.core.Either.Left
 import arrow.core.Either.Right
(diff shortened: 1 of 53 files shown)


raulraja (approved):
Thanks @nomisRev ! Looks great and painful.

rachelcarmena (approved):
:tada:

PR #326, opened 2021-02-04
Deprecate Order

nomisRev:
This PR deprecates `Order` since Kotlin, and Kotlin's Std doesn't take `Order` into account nor does Kotlin have the power to automatically derive or inject `Order` anywhere. This makes `Order` very cumbersome to work with, see `Tuple10Order.kt`

Another issue with `Order` is that non of the data types, or collections out there take `Order` into account. For example `SortedMap` doesn't work with `Order`, but still relies on the existing `Comparable` or `Comparator` type.

Sadly when implementing `operator fun A.compareTo` it still doesn't make the type `A` `Comparable<A>`.
It's easy to derive `Comparator`  in case all involved types are `Comparable<A>` for example, `val x: Comparator<Either<Int, Stringn>> = Comparator { a, b -> a.compareTo(b) }`.

Usage examples:

```kotlin
2.right() <= 3.right() // <=, and other comparing symbols enabled through `compareTo`

val x: List<Either<String, Int>> = listOf(3.right(), 1.right(), 2.right())
x.sortedWith(Either<String, Int>::compareTo) // Derive `Comparator` from `compareTo`

maxOf(2.right(), 3.right(), Either<String, Int>::compareTo)
minOf(2.right(), 3.right(), Either<String, Int>::compareTo)
sort(2.right(), 3.right(), 1,right(), -5.right(), comparator = Either<String, Int>::compareTo)
...
```

We could also expose `Either.comparator` from `Either`'s companion but it's almost as easy as just using a lambda reference to derive it from` compareTo` since `Comparator` is a `@FunctionalInterface` interface.


arrow-core-data/src/main/kotlin/arrow/core/Ordering.kt
```diff
@@ -3,13 +3,13 @@ package arrow.core
...
 import arrow.typeclasses.Hash
 import arrow.typeclasses.Monoid
-import arrow.typeclasses.Order
 import arrow.typeclasses.Semigroup
 import arrow.typeclasses.ShowDeprecation
 import arrow.typeclasses.hashWithSalt
 
-sealed class Ordering {
-  override fun equals(other: Any?): Boolean = this === other // ref equality is fine because objects should be singletons
+sealed class Ordering(private val toInt: Int) {
```
franciscodr, line 10: 👌

franciscodr (approved):
Thanks, @nomisRev!

raulraja (approved):
thanks @nomisRev !

arrow-core-data/src/main/kotlin/arrow/core/Tuple4.kt
```diff
@@ -117,33 +117,14 @@ fun <A, B, C, D> Hash.Companion.tuple4(
...
-
-private class Tuple4Order<A, B, C, D>(
-  private val OA: Order<A>,
-  private val OB: Order<B>,
-  private val OC: Order<C>,
-  private val OD: Order<D>
-) : Order<Tuple4<A, B, C, D>> {
-  override fun Tuple4<A, B, C, D>.compare(other: Tuple4<A, B, C, D>): Ordering =
-    compare(OA, OB, OC, OD, other)
+operator fun <A : Comparable<A>, B : Comparable<B>, C : Comparable<C>, D : Comparable<D>> Tuple4<A, B, C, D>.compareTo(other: Tuple4<A, B, C, D>): Int {
```
raulraja, line 120: Much nicer

arrow-core-data/src/main/kotlin/arrow/core/Ordering.kt
nomisRev: I'm wondering if we still need this type, or we should deprecate and remove it alongside `Order`.
You could use it in combination with `compareTo` to get a typed API instead of an `Int` based one.

```kotlin
val ordering = 2.right().compareTo(3.right())
  .let(Ordering.Companion::fromInt)
  
when(ordering) {
  is LT -> ...
  is GT -> ...
  is EQ -> ...
}
```

cc\\ @raulraja

nomisRev:
Turned this back into a draft since I need to update the already done deprecations for `Order`.

arrow-core-data/src/main/kotlin/arrow/core/Ordering.kt
raulraja: I think it's a useful utility if you don't want to deal with integers and want a typed api but in practice the purpose of the `compareTo` operator I interpret as a way for users to already syntactically enjoy `<`, `>`, and `==` which already covers all Boolean path cases those types cover. After all those types don't hold any data either. I personally think Ordering has no value but also I think it's harmless o happy to keep it as utility if anyone cares enough.

raulraja (approved):

franciscodr (approved):

PR #331, opened 2021-02-08
Make memoization compatible with MPP

aballano:
Replace ConcurrentHashMap impl by atomic to be able to work in MPP


arrow-syntax/src/main/kotlin/arrow/syntax/function/memoization.kt
```diff
@@ -214,6 +215,24 @@ private data class MemoizeKey22<out P1, out P2, out P3, out P4, out P5, out P6,
...
+      val b = k(f)
+      cache.loop { old ->
+        val bb = old[k]
+        // No value is set yet (race condition)
+        if (bb == null) {
+          // Value was set, return the value
+          if (cache.compareAndSet(old, old + Pair(k, b))) return b
+          // keep looping failed to set
+          else Unit
+          // A value was already set first, return it
```
nomisRev, line 232: We should add some concurrent tests to check for this branching.

aballano: Changes done in https://github.com/arrow-kt/arrow-core/pull/331/commits/0dbc7145665524c206f7daf7b3dfead30344dbea 

I added some docs and this single test

nomisRev: 👌

nomisRev (approved):
LGTM 👌 Thanks @aballano!

arrow-syntax/src/main/kotlin/arrow/syntax/function/memoization.kt
```diff
@@ -1,117 +1,208 @@
...
+ *   val value2 = memoizedF(42)
+ *
+ *   //sampleEnd
+ *   println("$value1 $value2")
+ * }
+ * ```
+ *
+ * Note that calling this function with the same parameters in parallel might cause the function to be executed twice.
+ */
 fun <R> (() -> R).memoize(): () -> R = object : () -> R {
```
i-walker, line 30: Does it make sense to define them as suspended functions. Given the nature of memoizing, or am I mistaken here? 
cc @nomisRev

nomisRev: This PR just updates the existing code to remove the dependency on Java classes. We can do a variant of this for suspending code as well.
Kinda silly that this needs to be defined separately tho, since normal code should work for suspending code. Only the other way around is not true.

nomisRev: Actually, the `suspend` version is the only pure/referential transparent implementation since it involves allocation mutable shared state to do the memoisation.

aballano:
@rachelcarmena @nomisRev Seems this PR is failing for arrow-fx due to test flakiness(?) or might have my changes affected that?

nomisRev:
@aballano seems to be the case of something flaky in Arrow Fx and Arrow Fx RxJava which are both deprecated so not very concerning but I've never seen the flaky test in Arrow Fx RxJava.

I restarted the build. Have you seen this multiple times on this PR?

aballano:
@nomisRev Yes, the previous build after merging master also failed due to failed test for Bracket in the stream package

arrow-syntax/src/main/kotlin/arrow/syntax/function/memoization.kt
i-walker: Yes, that was one of my concerns. 
Maybe it's worth deprecating it first and rolling out a suspended version either here or in `Arrow Fx`

i-walker (approved):
Thanks @aballano 🙌

PR #332, opened 2021-02-08
Deprecate use of Tuple2 and Tuple3

franciscodr:


arrow-core-data/src/main/kotlin/arrow/core/NonEmptyList.kt
```diff
@@ -331,8 +331,9 @@ class NonEmptyList<out A>(
       NonEmptyList(head.combine(b.head), tail.salign(SA, b.tail).toList())
     }
 
-  fun <B> padZip(other: NonEmptyList<B>): NonEmptyList<Tuple2<A?, B?>> =
-    NonEmptyList(Tuple2(head, other.head), tail.padZip(other.tail))
+  @JvmName("padZipTupled")
```
nomisRev, line 334: Why do we need this `@JvmName`?

arrow-core-data/src/main/kotlin/arrow/core/Tuple10.kt
```diff
@@ -3,18 +3,37 @@
...
+import arrow.KindDeprecation
 import arrow.typeclasses.Hash
 import arrow.typeclasses.Show
 import arrow.typeclasses.ShowDeprecation
 import arrow.typeclasses.defaultSalt
 
+@Deprecated(
+  message = KindDeprecation,
+  level = DeprecationLevel.WARNING
+)
```
nomisRev, line 15: 🙈

nomisRev (approved):
Looks great @franciscodr! 🥇 🥳

arrow-core-data/src/main/kotlin/arrow/core/NonEmptyList.kt
franciscodr: No, we don't need it. I think I put it when I was keeping the version of `padZip` with tuples, but we can get rid of it now.

rachelcarmena (approved):
:tada:

PR #333, opened 2021-02-08
Revert remove macros

danimontoya:


danimontoya:
Since we will not show those hierarchies anymore.. they can be removed. Closing this

PR #335, opened 2021-02-09
Fix typo in deprecation messages for Const data type

franciscodr:
There was a typo in several deprecation messages for the `Const` data type, where we were using `Cons` instead of `Const`.


raulraja (approved):
thanks @franciscodr !

danimontoya (approved):
Thank you @franciscodr 🙏 ❤️ !!!

nomisRev (approved):
Thanks Paco!!!

PR #337, opened 2021-02-09
Deprecate Hash

nomisRev:
This PR deprecates `Hash` since Kotlin, and Kotlin's Std doesn't take `Hash` into account for example for `HashMap` or `HashSet` nor does Kotlin have the power to automatically derive or inject `Hash` anywhere. This makes `Hash` very cumbersome to work with, see `Tuple10Hash.kt`

The use-cases of `Hash` seemed very limited and niche to us, so we've decided to exclude it from Arrow in 1.0.0. The common technique for this is to use data class, which automatically implements `hashCode()`.


franciscodr (approved):
Thanks @nomisRev!

arrow-core/src/main/kotlin/arrow/core/extensions/ordering/hash/OrderingHash.kt
```diff
@@ -30,5 +31,5 @@ fun Ordering.hashWithSalt(arg1: Int): Int = arrow.core.Ordering.hash().run {
   "UNCHECKED_CAST",
   "NOTHING_TO_INLINE"
 )
-@Deprecated("@extension kinded projected functions are deprecated", ReplaceWith("Hash.ordering()", "arrow.core.Hash", "arrow.core.ordering"))
+@Deprecated(OrderingDeprecation)
```
danimontoya, line 34: Here should we use `HashDeprecation` or `OrderingDeprecation` ?

danimontoya (approved):
Thanks @nomisRev 🙌 🔥 🔥

arrow-core/src/main/kotlin/arrow/core/extensions/ordering/hash/OrderingHash.kt
nomisRev: It's both derpecated so.. 😅 I preferred `Ordering` here since the type in itself is deprecated, so by extension all behavior for it is also deprecated.

danimontoya: perfect 👍 👍

Issue #20, opened 2020-02-20
Extract extensions properties for nested ADT values in core and data


raulraja:
Many nested values have their own identifiers:
```
Some.t
Left.l
Right.r
...
```
This issue is to track the creation of extension syntax for all those datatypes that bring a unified access to the inner property through property extension. 
Ex taken from slack where @hastebrot proposed:
```kotlin
package arrow.syntax.option

inline val <T> Some<T>.value get() = this.t
```
All these property should live in `arrow.syntax.[datatype]` for consistency.

PR #338, opened 2021-02-09
Unify syntax for extracting inner values

nomisRev:
Related to an old ticket which we want to clear up while moving towards 1.0.0

https://github.com/arrow-kt/arrow-core/issues/20

> Many nested values have their own identifiers:
>```
>Some.t
>Left.l
>Right.r
>...
>```
>
> This issue is to track the creation of extension syntax for all those datatypes that bring a unified access to the inner property through property extension.
> Ex taken from slack where @hastebrot proposed:

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
@@ -1067,7 +1067,11 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
    * The left side of the disjoint union, as opposed to the [Right] side.
    */
   @Suppress("DataClassPrivateConstructor")
-  data class Left<out A> @PublishedApi internal constructor(val a: A) : Either<A, Nothing>() {
+  data class Left<out A> @PublishedApi internal constructor(
+    @Deprecated("Use value instead", ReplaceWith("value"))
+    val a: A
+  ) : Either<A, Nothing>() {
+    val value: A = a
     override val isLeft = true
     override val isRight = false
 
@@ -1082,7 +1086,11 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
    * The right side of the disjoint union, as opposed to the [Left] side.
    */
   @Suppress("DataClassPrivateConstructor")
-  data class Right<out B> @PublishedApi internal constructor(val b: B) : Either<Nothing, B>() {
+  data class Right<out B> @PublishedApi internal constructor(
+    @Deprecated("Use value instead", ReplaceWith("value"))
+    val b: B
+  ) : Either<Nothing, B>() {
+    val value: B = b
     override val isLeft = false
     override val isRight = true
 
--- a/arrow-core-data/src/main/kotlin/arrow/core/Tuple10.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Tuple10.kt
@@ -33,7 +33,7 @@ data class Tuple10<out A, out B, out C, out D, out E, out F, out G, out H, out I
   val f: F,
   @Deprecated("Use seventh instead", ReplaceWith("seventh"))
   val g: G,
-  @Deprecated("Use eigth instead", ReplaceWith("eigth"))
+  @Deprecated("Use eighth instead", ReplaceWith("eighth"))
   val h: H,
   @Deprecated("Use ninth instead", ReplaceWith("ninth"))
   val i: I,
@@ -166,12 +166,12 @@ operator fun <A : Comparable<A>, B : Comparable<B>, C : Comparable<C>, D : Compa
             if (sixth == 0) {
               val seventh = g.compareTo(other.g)
               if (seventh == 0) {
-                val eigth = h.compareTo(other.h)
-                if (eigth == 0) {
+                val eighth = h.compareTo(other.h)
+                if (eighth == 0) {
                   val ninth = i.compareTo(other.i)
                   if (ninth == 0) j.compareTo(other.j)
                   else ninth
-                } else eigth
+                } else eighth
               } else seventh
             } else sixth
           } else fifth
--- a/arrow-core-data/src/main/kotlin/arrow/core/Tuple10.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Tuple10.kt
@@ -18,7 +18,39 @@ typealias Tuple10PartialOf<A, B, C, D, E, F, G, H, I> = arrow.Kind9<ForTuple10, 
 inline fun <A, B, C, D, E, F, G, H, I, J> Tuple10Of<A, B, C, D, E, F, G, H, I, J>.fix(): Tuple10<A, B, C, D, E, F, G, H, I, J> =
   this as Tuple10<A, B, C, D, E, F, G, H, I, J>
 
-data class Tuple10<out A, out B, out C, out D, out E, out F, out G, out H, out I, out J>(val a: A, val b: B, val c: C, val d: D, val e: E, val f: F, val g: G, val h: H, val i: I, val j: J) : Tuple10Of<A, B, C, D, E, F, G, H, I, J> {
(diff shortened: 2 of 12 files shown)


arrow-core-data/src/main/kotlin/arrow/core/Tuple9.kt
```diff
@@ -18,7 +18,37 @@ typealias Tuple9PartialOf<A, B, C, D, E, F, G, H> = arrow.Kind8<ForTuple9, A, B,
...
+) : Tuple9Of<A, B, C, D, E, F, G, H, I> {
+
+  val first: A = a
+  val second: B = b
+  val third: C = c
+  val fourth: D = d
+  val fifth: E = e
+  val sixth: F = f
+  val seventh: G = g
+  val eight: H = h
```
franciscodr, line 49:
```suggestion
  val eighth: H = h
```

arrow-core-data/src/main/kotlin/arrow/core/Tuple9.kt
```diff
@@ -18,7 +18,37 @@ typealias Tuple9PartialOf<A, B, C, D, E, F, G, H> = arrow.Kind8<ForTuple9, A, B,
...
+  val c: C,
+  @Deprecated("Use fourth instead", ReplaceWith("fourth"))
+  val d: D,
+  @Deprecated("Use fifth instead", ReplaceWith("fifth"))
+  val e: E,
+  @Deprecated("Use sixth instead", ReplaceWith("sixth"))
+  val f: F,
+  @Deprecated("Use seventh instead", ReplaceWith("seventh"))
+  val g: G,
+  @Deprecated("Use eigth instead", ReplaceWith("eigth"))
```
franciscodr, line 36:
```suggestion
  @Deprecated("Use eigth instead", ReplaceWith("eighth"))
```

arrow-core-data/src/main/kotlin/arrow/core/Tuple8.kt
```diff
@@ -18,7 +18,34 @@ typealias Tuple8PartialOf<A, B, C, D, E, F, G> = arrow.Kind7<ForTuple8, A, B, C,
...
+  val c: C,
+  @Deprecated("Use fourth instead", ReplaceWith("fourth"))
+  val d: D,
+  @Deprecated("Use fifth instead", ReplaceWith("fifth"))
+  val e: E,
+  @Deprecated("Use sixth instead", ReplaceWith("sixth"))
+  val f: F,
+  @Deprecated("Use seventh instead", ReplaceWith("seventh"))
+  val g: G,
+  @Deprecated("Use eigth instead", ReplaceWith("eigth"))
```
franciscodr, line 36:
```suggestion
  @Deprecated("Use eigth instead", ReplaceWith("eighth"))
```

arrow-core-data/src/main/kotlin/arrow/core/Tuple8.kt
```diff
@@ -18,7 +18,34 @@ typealias Tuple8PartialOf<A, B, C, D, E, F, G> = arrow.Kind7<ForTuple8, A, B, C,
...
+) : Tuple8Of<A, B, C, D, E, F, G, H> {
+
+  val first: A = a
+  val second: B = b
+  val third: C = c
+  val fourth: D = d
+  val fifth: E = e
+  val sixth: F = f
+  val seventh: G = g
+  val eight: H = h
```
franciscodr, line 47:
```suggestion
  val eighth: H = h
```

arrow-core-data/src/main/kotlin/arrow/core/Tuple10.kt
```diff
@@ -18,7 +18,39 @@ typealias Tuple10PartialOf<A, B, C, D, E, F, G, H, I> = arrow.Kind9<ForTuple10,
...
+  val c: C,
+  @Deprecated("Use fourth instead", ReplaceWith("fourth"))
+  val d: D,
+  @Deprecated("Use fifth instead", ReplaceWith("fifth"))
+  val e: E,
+  @Deprecated("Use sixth instead", ReplaceWith("sixth"))
+  val f: F,
+  @Deprecated("Use seventh instead", ReplaceWith("seventh"))
+  val g: G,
+  @Deprecated("Use eigth instead", ReplaceWith("eigth"))
```
franciscodr, line 36:
```suggestion
  @Deprecated("Use eigth instead", ReplaceWith("eighth"))
```

arrow-core-data/src/main/kotlin/arrow/core/Tuple10.kt
```diff
@@ -18,7 +18,39 @@ typealias Tuple10PartialOf<A, B, C, D, E, F, G, H, I> = arrow.Kind9<ForTuple10,
...
+) : Tuple10Of<A, B, C, D, E, F, G, H, I, J> {
+
+  val first: A = a
+  val second: B = b
+  val third: C = c
+  val fourth: D = d
+  val fifth: E = e
+  val sixth: F = f
+  val seventh: G = g
+  val eight: H = h
```
franciscodr, line 51:
```suggestion
  val eighth: H = h
```

arrow-core-data/src/main/kotlin/arrow/core/Tuple9.kt
aballano: I think we should apply the wording fix to the message too "Use eighth instead" (here and below)

nomisRev: Fixed in both places

arrow-core-data/src/main/kotlin/arrow/core/Tuple8.kt
```diff
@@ -18,7 +18,34 @@ typealias Tuple8PartialOf<A, B, C, D, E, F, G> = arrow.Kind7<ForTuple8, A, B, C,
...
+  val c: C,
+  @Deprecated("Use fourth instead", ReplaceWith("fourth"))
+  val d: D,
+  @Deprecated("Use fifth instead", ReplaceWith("fifth"))
+  val e: E,
+  @Deprecated("Use sixth instead", ReplaceWith("sixth"))
+  val f: F,
+  @Deprecated("Use seventh instead", ReplaceWith("seventh"))
+  val g: G,
+  @Deprecated("Use eighth instead", ReplaceWith("eigth"))
```
aballano, line 36:
```suggestion
  @Deprecated("Use eighth instead", ReplaceWith("eighth"))
```

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -1067,7 +1067,11 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
    * The left side of the disjoint union, as opposed to the [Right] side.
    */
   @Suppress("DataClassPrivateConstructor")
-  data class Left<out A> @PublishedApi internal constructor(val a: A) : Either<A, Nothing>() {
+  data class Left<out A> @PublishedApi internal constructor(
```
raulraja, line 1070: Would making the constructor internal break user code at call sites like `Left(1)`?
Can we do instead something like?:
```kotlin
data class Left<out A> @PublishedApi internal constructor(@Deprecated("use value") val a: A, val value : A = a)
```

aballano (approved):

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
raulraja: Nevermind, just realized it was like that before

raulraja (approved):
Thanks @nomisRev !

arrow-core-data/src/main/kotlin/arrow/core/Either.kt
nomisRev: The constructor is already internal. I'm not sure what the difference would be 🤔

franciscodr (approved):
Thanks @nomisRev!

PR #341, opened 2021-02-11
CU-bezf1h deprecate Arrow Annotations

nomisRev:
This PR deprecates some of the annotations in Arrow that will no longer be supported, most of them relating to higher kinded types.

### DIFF
--- a/arrow-annotations/src/main/java/arrow/synthetic.kt
+++ b/arrow-annotations/src/main/java/arrow/synthetic.kt
@@ -31,5 +31,4 @@ package arrow
   AnnotationTarget.TYPEALIAS
 )
 @MustBeDocumented
-@Deprecated("This annotation is deprecated along with the deprecation of the Arrow-meta kapt implementation.")
 annotation class synthetic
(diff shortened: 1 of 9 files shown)


franciscodr:
Task linked: [CU-bezf1h Deprecate Arrow Annotations](https://app.clickup.com/t/bezf1h)

arrow-annotations/src/main/java/arrow/synthetic.kt
```diff
@@ -31,4 +31,5 @@ package arrow
   AnnotationTarget.TYPEALIAS
 )
 @MustBeDocumented
+@Deprecated("This annotation is deprecated along with the deprecation of the Arrow-meta kapt implementation.")
```
raulraja, line 34: This one is actually used by Arrow Meta (upcoming) not the kapt version.
```suggestion
```

raulraja (approved):
thanks @nomisRev !

arrow-annotations/src/main/java/arrow/synthetic.kt
nomisRev: Oh okay, I missunderstood you before then 😅 Thanks 😉

i-walker (approved):
Thanks @nomisRev

PR #344, opened 2021-02-12
Clean up constructors Arrow Core Datatypes.

nomisRev:
This PR deprecates up all duplicated constructors, and deprecates all `just` constructors.

This leaves behind only the actual constructors, and the extension constructors.

All constructors of Either after removal of deprecated methods:

```
val x: Either<String, Int> = Either.Right(1)
val y: Either<String, Int> = Either.Left("one")
val x2: Either<String, Int> = 1.right()
val y2: Either<String, Int> = "one".left()
```

### DIFF
--- a/arrow-core-data/src/main/kotlin/arrow/core/Eval.kt
+++ b/arrow-core-data/src/main/kotlin/arrow/core/Eval.kt
@@ -183,7 +183,7 @@ sealed class Eval<out A> : EvalOf<A> {
     fun raise(t: Throwable): Eval<Nothing> =
       defer { throw t }
 
-    val unit: Eval<Unit> = Now(kotlin.Unit)
+    val Unit: Eval<Unit> = Now(kotlin.Unit)
 
     val True: Eval<Boolean> = Now(true)
 
@@ -284,7 +284,7 @@ sealed class Eval<out A> : EvalOf<A> {
       b: Eval<B>,
       map: (A, B) -> C
     ): Eval<C> =
-      mapN(a, b, unit, unit, unit, unit, unit, unit, unit) { aa, bb, _, _, _, _, _, _, _ -> map(aa, bb) }
+      mapN(a, b, Unit, Unit, Unit, Unit, Unit, Unit, Unit) { aa, bb, _, _, _, _, _, _, _ -> map(aa, bb) }
 
     fun <A, B, C, D> mapN(
       a: Eval<A>,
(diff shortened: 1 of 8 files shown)


franciscodr (approved):
Thanks @nomisRev!

arrow-core-data/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -431,6 +431,14 @@ sealed class Option<out A> : OptionOf<A> {
...
+      "empty is deprecated, and will be removed in 0.13.0. Please use None instead.",
+      ReplaceWith(
+        "None",
+        "arrow.core.None"
+      ),
+      DeprecationLevel.WARNING
+    )
     fun <A> empty(): Option<A> = None
 
     val unit: Option<Unit> = Some(Unit)
```
franciscodr, line 444: I think we also discussed removing the `unit` constructor method, didn't we?

nomisRev: Yes, you're correct @franciscodr. I updated the PR. I checked against 0.11.0.3 and most of these were added after, so we could get away with removing them insteaad of deprecating 😄

Thanks for the reminder! 🕵️

arrow-core-data/src/main/kotlin/arrow/core/Eval.kt
```diff
@@ -183,7 +183,7 @@ sealed class Eval<out A> : EvalOf<A> {
     fun raise(t: Throwable): Eval<Nothing> =
       defer { throw t }
 
-    val unit: Eval<Unit> = Now(kotlin.Unit)
+    val Unit: Eval<Unit> = Now(kotlin.Unit)
```
nomisRev, line 186: This was uppercased on `0.11.0`

arrow-core-data/src/main/kotlin/arrow/core/NonEmptyList.kt
```diff
@@ -346,10 +356,19 @@ class NonEmptyList<out A>(
     fun <A> fromListUnsafe(l: List<A>): NonEmptyList<A> =
       NonEmptyList(l)
 
+    @Deprecated(
+      "just is deprecated, and will be removed in 0.13.0. Please use NonEmptyList.of instead.",
+      ReplaceWith(
+        "NonEmptyList.of(a)",
```
nomisRev, line 362:
```suggestion
        "nonEmptyListOf(a)",
```

arrow-core-data/src/main/kotlin/arrow/core/Eval.kt
i-walker: Do we want those to be upper-cased or lowercased in Arrow for other libs and API's?

i-walker (approved):
Thanks @nomisRev 🙌 , left a general question below on naming conventions.

arrow-core-data/src/main/kotlin/arrow/core/Eval.kt
nomisRev: Looking at the other (smart) constructors in Arrow, I think we should deprecate all of these in favor of their right hand side implementations. 

```
val Unit: Eval<Unit> = Now(kotlin.Unit)
val True: Eval<Boolean> = Now(true)
val False: Eval<Boolean> = Now(false)
val Zero: Eval<Int> = Now(0)
val One: Eval<Int> = Now(1)
```

If all agree I can add this change in this PR as well.

i-walker: I am sorry, I saw the reply after merging it. 
Let's move that to a different PR 😄

